PR #151, opened 2026-01-15
feat(graph): add valid IDs context for serialize phase

pvliesdonk:
## Problem
LLMs generating SEED output don't know what entity/tension IDs are valid until semantic validation fails. This leads to:
- Phantom ID references (referencing entities that don't exist)
- Wasted retries fixing validation errors
- Poor LLM experience - they're flying blind

## Changes
- Add `src/questfoundry/graph/context.py` with `format_valid_ids_context()` function
- Inject valid IDs context into brief at start of `serialize_seed_iteratively()`
- Export `format_valid_ids_context` from graph package
- Add 9 comprehensive tests for context formatting

The context includes:
- Entity IDs grouped by category (character, location, object, faction)
- Tension IDs with their alternative IDs (marking default path)
- Rules for ID usage in SEED stage

## Not Included / Future PRs
- GROW stage valid IDs (deferred until GROW is implemented)
- PR 4: Orchestrator cleanup + post-mutation checks
- PR 5: Update tests using deprecated methods

## Test Plan
```bash
uv run pytest tests/unit/test_graph_context.py -v  # 9 passed
uv run pytest tests/unit/ -q  # 604 passed
```

## Risk / Rollback
- Low risk - additive change that doesn't modify existing behavior
- If context is too verbose, can adjust formatting
- Rollback: remove context injection from serialize.py

### DIFF
--- a/src/questfoundry/agents/serialize.py
+++ b/src/questfoundry/agents/serialize.py
@@ -518,10 +518,11 @@ async def serialize_seed_iteratively(
 
     # Inject valid IDs context if graph is provided
     # This gives the LLM authoritative ID list upfront to prevent phantom references
+    enhanced_brief = brief
     if graph is not None:
         valid_ids_context = format_valid_ids_context(graph, stage="seed")
         if valid_ids_context:
-            brief = f"{valid_ids_context}\n\n---\n\n{brief}"
+            enhanced_brief = f"{valid_ids_context}\n\n---\n\n{brief}"
             log.debug("valid_ids_context_injected", context_length=len(valid_ids_context))
 
     # Section configuration: (section_name, schema, output_field)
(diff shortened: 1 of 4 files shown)


pvliesdonk:
## Review Comments Addressed

Addressed all review feedback in commit 183babb:

**@gemini-code-assist - Performance optimization (context.py:105)**
✅ Fixed: Pre-built alt edges map for O(T+E) performance instead of O(T*E). Also added `alts.sort()` for deterministic output as suggested.

**@claude - defaultdict suggestion (context.py:60)**
✅ Acknowledged: Minor suggestion, current code with `setdefault()` is equivalent and works fine. Keeping as-is since it's not a blocker.

**@claude - brief mutation question (serialize.py)**
✅ Fixed: Created `enhanced_brief` variable instead of mutating the `brief` parameter. This makes the modification explicit and local to the function.

PR #185, opened 2026-01-17
chore: update models and ifcraftcorpus dependency

pvliesdonk:
## Problem

QoL updates to keep dependencies and model defaults current:
- ifcraftcorpus 1.2.1 doesn't support FTS5 syntax
- Ollama tests using older qwen3:8b model
- OpenAI tests/defaults using gpt-4o instead of newer gpt-5-mini

## Changes

- **ifcraftcorpus**: Updated to 1.4.0 (adds FTS5 syntax support)
- **Ollama model**: Changed default/test model to `qwen3:4b-instruct-32k`
- **OpenAI model**: Changed default/test model to `gpt-5-mini` (1M context window)
- Updated all tests, examples, and documentation to use new model names

## Not Included / Future PRs

- No functional changes to pipeline behavior

## Test Plan

- [x] `uv run pytest tests/unit/` - 704 tests pass
- [x] All model info tests updated for new context window sizes

## Risk / Rollback

Low risk - only changes model names in defaults/tests and updates a dependency version.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #197, opened 2026-01-18
fix: remove broken entity coverage validation from SEED stage

pvliesdonk:
## Problem

The `validate_entity_coverage` implementation was fundamentally flawed - a category error:

1. It validated **format** (regex matching `id: entity_id`) instead of **semantics**
2. LLMs produce markdown tables, not the expected format
3. Validation found 0 matches in all runs, triggering useless retries

Evidence from projects/seq-{1..6}:
```
summarize_validation_failed, attempt=1, expected=17, actual=0
summarize_validation_failed, attempt=2, expected=17, actual=0
```

## Root Cause Analysis

Format validation on free-text summarize output is architecturally wrong. The correct approach:
- Let summarize produce any format the LLM chooses
- Serialize phase validates semantics against the graph
- `serialize_with_brief_repair` already implements the correct two-level loop

## Changes

- Remove `validate_entity_coverage`, `_extract_entity_checklist`, `_count_tensions`, `get_expected_entity_count` from prompts.py
- Remove broken validation params from SEED stage summarize call
- Remove `import re` (no longer needed)
- Remove tests for removed functions (-21 tests)

## Not Included

- Kept the optional `entity_validator` param in `summarize_discussion` signature (dormant, may be useful for other stages)
- No changes to `serialize_with_brief_repair` - it's the correct architecture

## Test Plan

```bash
uv run pytest tests/unit/ -q  # 728 passed (was 749, removed 21)
```

## Architectural Note

The existing two-level loop in `serialize_with_brief_repair` is correct:
- Inner loop: Per-section Pydantic validation
- Outer loop: Brief repair on semantic failure (invalid IDs)

No additional loop for summarize regeneration is needed - that would violate the "bounded iteration" principle.


pvliesdonk:
Closing as part of rollback to #185. All SEED two-level feedback loop changes (#188-#201) are being reverted in PR #203.

The root cause (phantom ID invention) will be addressed differently in issue #202 by splitting SEED into 4 focused phases.

Pre-revert code preserved in branch `backup/pre-revert-to-185`.

PR #289, opened 2026-01-24
feat(grow): add observability with @traceable and token extraction (#282)

### LINKED ISSUE #282: fix(grow): Add token tracking and @traceable observability
## Problem

GROW lacks two key observability features that other stages have:

### 1. Token tracking always returns 0

`_grow_llm_call()` at `grow.py:326` always returns `0` for token count:
```python
return result, llm_calls, 0  # tokens always 0
```

This means `total_tokens` in `execute()` is meaningless, and the completion log at line 252 never reports real usage. The established stages extract tokens via `_extract_tokens()` from response `usage_metadata`.

### 2. No @traceable decorators

DREAM/BRAINSTORM/SEED wrap key functions with `@traceable` for LangSmith tracing:
- `run_discuss_phase()` → traced
- `summarize_discussion()` → traced  
- `serialize_to_artifact()` → traced

GROW has no `@traceable` on:
- `execute()`
- `_grow_llm_call()`
- Individual phase methods (`_phase_2_thread_agnostic`, etc.)

This means GROW phases are invisible in LangSmith traces.

### Token tracking
Extract tokens from the structured model response in `_grow_llm_call()` using the same pattern as `agents/serialize.py`:
```python
tokens = 0
if hasattr(result, "response_metadata"):
    usage = result.response_metadata.get("usage", {})
    tokens = usage.get("total_tokens", 0)
```

Or use the `_extract_tokens()` helper if it can be made importable.

### Tracing
Add `@traceable` to:
- `execute()` (stage-level trace)
- `_grow_llm_call()` (per-call trace with template name and phase context)

## Impact

Without this fix:
- Cannot track LLM costs for GROW stages
- Cannot debug GROW phases in LangSmith
- `--log` output is incomplete for GROW

pvliesdonk:
## Problem
GROW stage LLM calls always return `tokens_used=0`, making it impossible to track token consumption. Additionally, the stage lacks tracing support for LangSmith observability.

## Changes
- Add `@traceable` decorator to `execute()` and `_grow_llm_call()` for LangSmith tracing
- Import and use `_extract_tokens()` from serialize module after each `ainvoke()` call
- Track `total_tokens` across retry attempts within `_grow_llm_call()`
- Return actual accumulated token counts instead of hardcoded `0`
- Update `execute()` docstring to reflect correct return semantics

## Not Included / Future PRs
- Per-phase token breakdowns in log output (can be added as follow-up)
- Custom metadata injection per phase template (GROW phases share the same traceable config)

## Test Plan
- `test_tokens_extracted_from_result`: Mocks `_extract_tokens` to return 150, verifies propagation
- `test_tokens_zero_when_no_metadata`: Verifies 0 returned when no usage metadata available
- `test_tokens_accumulate_across_retries`: Verifies tokens sum across retry attempts (100+120=220)
- All 79 existing GROW tests pass (unit + integration)

```bash
uv run pytest tests/unit/test_grow_stage.py tests/integration/test_grow_e2e.py -xvs
uv run ruff check src/questfoundry/pipeline/stages/grow.py
```

## Risk / Rollback
- `@traceable` is a no-op without `LANGSMITH_TRACING=true` — zero runtime impact when tracing disabled
- `_extract_tokens` returns 0 for structured output results (Pydantic instances lack metadata), so behavior matches previous `0` for most providers until LangChain passes metadata through

Closes #282

### DIFF
--- a/src/questfoundry/agents/serialize.py
+++ b/src/questfoundry/agents/serialize.py
@@ -266,7 +266,7 @@ async def serialize_to_artifact(
                 result = await structured_model.ainvoke(messages, config=config)
 
                 # Extract token usage from response if available
-                tokens = _extract_tokens(result)
+                tokens = extract_tokens(result)
                 total_tokens += tokens
 
                 # If result is already a Pydantic model, validate succeeded
@@ -384,7 +384,7 @@ async def serialize_to_artifact(
     )
 
 
-def _extract_tokens(result: object) -> int:
+def extract_tokens(result: object) -> int:
     """Extract token usage from response metadata.
 
     LangChain tracks token usage in different places:
--- a/src/questfoundry/pipeline/stages/grow.py
+++ b/src/questfoundry/pipeline/stages/grow.py
@@ -26,11 +26,13 @@ from typing import TYPE_CHECKING, Any, TypeVar
 from langchain_core.messages import HumanMessage, SystemMessage
 from pydantic import BaseModel, ValidationError
 
+from questfoundry.agents.serialize import _extract_tokens
 from questfoundry.artifacts.validator import get_all_field_paths
 from questfoundry.graph.graph import Graph
 from questfoundry.graph.mutations import GrowMutationError, GrowValidationError
 from questfoundry.models.grow import GrowPhaseResult, GrowResult
 from questfoundry.observability.logging import get_logger
+from questfoundry.observability.tracing import traceable
 from questfoundry.pipeline.gates import AutoApprovePhaseGate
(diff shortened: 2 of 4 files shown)


pvliesdonk:
## Review feedback addressed in d36b15d

1. **Private function import**: Renamed `_extract_tokens` → `extract_tokens` across all 4 files (serialize.py, grow.py, and their tests). The function is now public API since it's used cross-module.

2. **Duplicated log+return**: Consolidated the two return paths (isinstance check + model_validate fallback) into a single ternary + shared log/return.

PR #308, opened 2026-01-25
fix(seed): set is_canonical on threads from alternative's is_default_path

pvliesdonk:
## Problem

When creating threads from tension alternatives, the `is_canonical` flag wasn't being
set based on the alternative's `is_default_path` property. This meant all threads
had `is_canonical=False` even when they represented the default path.

## Changes

- **mutations.py**: Copy `is_default_path` from alternative to thread's `is_canonical` during mutation processing
- Add test to verify this behavior

## Not Included / Future PRs

Part of larger refactor branch being split per PR guidelines.

## Test Plan

```bash
uv run pytest tests/unit/test_mutations.py -v -k "canonical"
# 1 passed
```

## Risk / Rollback

- Low risk - additive fix
- Affects downstream GROW phase which uses is_canonical for arc enumeration


pvliesdonk:
Closing - restructuring into stacked PRs at commit boundaries

PR #323, opened 2026-01-26
feat: rename explored to considered with derived state

pvliesdonk:
## Problem

The over-generate-and-select pattern has a bug where tensions with `explored >= 2` but only one actual thread slip through pruning. This happens because:
1. Arc count was computed from the `explored` field, not thread existence
2. Tension scoring used different criteria than arc counting
3. Pruning mutated tensions, making the field unreliable

## Changes

- **Rename `explored` → `considered`**: Better reflects semantic meaning (LLM's intent)
- **Add backward compat validator**: Old graphs with `explored` field auto-migrate
- **Derive development state from threads**: `committed`/`deferred`/`latent` computed at runtime
- **Remove tension mutation in pruning**: `considered` is now immutable after SEED
- **Fix arc count derivation**: Now counts threads per tension, not `considered` field
- **Fix scoring consistency**: `is_fully_explored` checks thread existence

### Key Invariants

1. `canonical ∈ committed` (spine path always has thread)
2. `committed ⊆ considered` (can only develop what was considered)
3. `arc_count = 2^n` where `n = tensions with 2+ threads`
4. Pruning only drops threads, never modifies tension fields

## Not Included / Future PRs

- Migration script for existing project graphs (backward compat handles reading)
- UI updates if any reference the old field name

## Test Plan

- [x] All 1322 existing tests pass
- [x] 15 new tests for ontology change:
  - Backward compatibility (old `explored` migrates to `considered`)
  - Development state derivation
  - Arc count from thread existence
  - Pruning immutability

```bash
uv run pytest  # 1322 passed, 6 xpassed
```

## Risk / Rollback

- Low risk: backward compatibility validator handles old graphs
- Rollback: revert PR, old graphs still readable

### DIFF
--- a/docs/design/00-spec.md
+++ b/docs/design/00-spec.md
@@ -336,10 +336,17 @@ tension:
   involves: entity_id[]
   why_it_matters: string        # thematic stakes
   # Added by SEED:
-  explored: alternative_id[]    # which alternatives become threads
+  considered: alternative_id[]  # which alternatives LLM intended to explore
 ```
 
 **Lifecycle:** Created in BRAINSTORM, exploration decisions added in SEED. Not exported.
+
+**Derived development states** (computed from thread existence, not stored):
+- **committed**: Alternative has a thread in the graph (will become a story path)
+- **deferred**: Alternative in `considered` but no thread (LLM intended to explore but was pruned)
+- **latent**: Alternative not in `considered` (never intended for exploration, becomes shadow)
+
+The `considered` field records what the LLM *intended* to explore. Actual thread existence determines what was *committed*. This separation allows pruning to drop threads without modifying the tension's stored intent, keeping the field immutable after SEED.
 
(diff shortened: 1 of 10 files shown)


pvliesdonk:
## Review Feedback Addressed ✅

Fixed in commit `889c69e`:

### 1. Fallback Logic Bug (mutations.py:1171, 1305)

Changed from:
```python
considered = tension_decision.get("considered") or tension_decision.get("explored", [])
```

To:
```python
considered = tension_decision.get("considered", tension_decision.get("explored", []))
```

This correctly handles empty lists (which are falsy but valid).

### 2. Documentation: Purpose of `implicit` field

Added documentation in `docs/design/00-spec.md`:

> **The `implicit` field** holds alternatives that are intentionally NOT explored as threads. These provide narrative context for the FILL stage—the "road not taken" that gives meaning to the chosen path.

---

Ready for re-review.

Issue #361, opened 2026-01-29
Enhancement: Split path-specific lead-ins for conditional prerequisites


pvliesdonk:
### Context

When `check_intersection_compatibility()` detects a conditional prerequisite (a shared beat depending on a path-specific beat), it currently **rejects the intersection** entirely. This is the safest strategy but may reduce graph connectivity.

This issue tracks an alternative: instead of rejecting, **split the shared beat** into path-specific variants so the version with the dependency keeps it while other paths get an independent copy.

### Proposed Behavior

When intersection of beats `[A, B]` would span paths `{P1, P2, P3, P4}`, and beat `A` requires beat `G` which only belongs to `{P1}`:

1. Create `A_p1` (copy of `A`) that keeps `requires → G`, belonging to `{P1}`
2. Keep original `A` for `{P2, P3, P4}` without the `requires → G` edge
3. Proceed with intersection of `[A, B]` for paths `{P2, P3, P4}` only
4. `A_p1` remains a separate beat on `{P1}`

### Tradeoffs

- **Pro:** Preserves the intersection for paths where it is safe
- **Pro:** Respects the narrative intent of the path-specific prerequisite
- **Con:** Increases beat count and graph complexity
- **Con:** The split beat needs its own prose in FILL (duplication)

### Code Location

`src/questfoundry/graph/grow_algorithms.py` → `check_intersection_compatibility()`, the conditional-prerequisite invariant check.

### Related

- #357: Original `passage_dag_cycles` failure
- #358: Phase reordering (gaps before intersections)
- #359: No-conditional-prerequisites invariant design
- #360: Alternative: lift prerequisites into shared set

Issue #360, opened 2026-01-29
Enhancement: Lift prerequisites into shared set during intersection


pvliesdonk:
### Context

When `check_intersection_compatibility()` detects a conditional prerequisite (a shared beat depending on a path-specific beat), it currently **rejects the intersection** entirely. This is the safest strategy but may reduce graph connectivity.

This issue tracks an alternative: instead of rejecting, **lift the prerequisite beat into the shared set** by adding `belongs_to` edges so it spans all intersection paths.

### Proposed Behavior

When intersection of beats `[A, B]` would span paths `{P1, P2, P3, P4}`, and beat `A` requires beat `G` which only belongs to `{P1}`:

1. Add `belongs_to` edges so `G` also belongs to `{P2, P3, P4}`
2. Proceed with the intersection as normal

### Tradeoffs

- **Pro:** Preserves the intersection, keeping the story graph more connected
- **Con:** May widen a beat beyond its narrative intent — a gap beat designed for one path's pacing now appears in all paths
- **Con:** May require the lifted beat's `requires` targets to also be lifted (transitive closure)

### Code Location

`src/questfoundry/graph/grow_algorithms.py` → `check_intersection_compatibility()`, the conditional-prerequisite invariant check.

### Related

- #357: Original `passage_dag_cycles` failure
- #358: Phase reordering (gaps before intersections)
- #359: No-conditional-prerequisites invariant design

PR #362, opened 2026-01-29
fix(grow): reorder phases and enforce no-conditional-prerequisites invariant

### LINKED ISSUE #357: Bug: GROW validation fails with passage DAG cycles caused by cross-arc ordering flips
**Bug Description**
GROW validation fails with `passage_dag_cycles` because the passage successor graph is built from *all* arcs. When arcs have different beat ordering (due to gap insertion + requires edges), shared passages can flip order between arcs, creating edges in both directions and a cycle.

**Evidence (projects/test, run at 2026-01-28 17:51)**
Validation failure:
```
GROW has validation errors:
  - validation: 1 failed, 4 warnings, 5 passed
```
Failing check:
```
passage_dag_cycles: Cycle detected involving 39 passages: passage::ai_casualty_or_misguided__casualty_beat_04,
 passage::archive_guardians_preserve_or_reveal__reveal_beat_01,
 passage::archive_guardians_preserve_or_reveal__reveal_beat_02,
 passage::archive_guardians_preserve_or_reveal__reveal_beat_03,
 passage::archive_guardians_preserve_or_reveal__reveal_beat_04
```
One concrete cycle:
```
passage::archive_guardians_preserve_or_reveal__reveal_beat_03
 -> passage::gap_1
 -> passage::ai_casualty_or_misguided__casualty_beat_04
 -> passage::gap_5
 -> passage::archive_guardians_preserve_or_reveal__reveal_beat_01
 -> passage::archive_guardians_preserve_or_reveal__reveal_beat_03
```
Choices generated for those edges confirm the cycle (phase 9):
- reveal_beat_03 → gap_1
- gap_1 → casualty_beat_04
- casualty_beat_04 → gap_5
- gap_5 → reveal_beat_01
- reveal_beat_01 → reveal_beat_03

**Root Cause (analysis)**
`find_passage_successors()` builds a passage→passage graph by iterating **all arcs**, deduplicating successors per passage. Because arcs are topologically sorted over a union of beats, shared passages can appear in different relative orders across arcs. When that happens, we end up with successors A→B in one arc and B→A in another, which creates a cycle in the global passage graph.

This is likely triggered by gap insertion (`beat::gap_*`) adding `requires` edges that reorder beats in some arcs but not others.

**Initial Assessment / Directions**
Potential fixes (choose one or combine):
1) **Canonical ordering**: Build passage successors from a single canonical arc (e.g., spine) instead of all arcs. This avoids cross-arc ordering flips but may reduce choice coverage.
2) **Global ordering constraint**: Enforce stable relative ordering for beats that share a path or intersection group, so all arcs preserve the same order for shared passages.
3) **Edge filtering**: When creating choices, drop successors that would move “backward” in spine order, preventing cycles at the cost of some branches.

**Files of interest**
- `src/questfoundry/graph/grow_algorithms.py` (`find_passage_successors`)
- `src/questfoundry/graph/grow_validation.py` (`check_passage_dag_cycles`)
- `src/questfoundry/pipeline/stages/grow.py` (phase 9 choices)

**Run Artifacts**
- `projects/test/snapshots/grow-pre-validation.json`
- `projects/test/logs/debug.jsonl`

### LINKED ISSUE #358: Refactor: Reorder GROW phases (Gaps before Intersections) to prevent topological paradoxes
### Refactoring Phase Order to Prevent Topological Paradoxes

**Context from #357**
We identified a critical structural flaw in the GROW stage:
1. **Phase 3 (Intersections)** merges beats from different paths (e.g., `Path A` and `Path B`) into a single "shared" beat.
2. **Phase 4 (Gaps)** inserts a new beat (`Gap 1`) into `Path A` as a prerequisite for that shared beat.
3. **The Paradox**: The shared beat now requires `Gap 1`. In `Path B`, `Gap 1` does not exist. This creates undefined ordering in `Path B` (the prerequisite is missing), leading to "floating" nodes and topological cycles (`passage_dag_cycles`) when arcs are enumerated.

**Diagnosis: Improper Procedure**
We are currently "weaving threads before they are finished." By calculating intersections (weaving) *before* filling narrative gaps (thread completion), we lock the graph structure into a shared state that might be incompatible with the detailed pacing requirements of individual paths.

**Proposed Solution: Reorder GROW Phases**
We should move **Gap Detection (Phase 4)** to run **before** **Intersection Detection (Phase 3)**.

**New Phase Order:**
1.  **Phase 1**: Validate DAG.
2.  **Phase 2**: Path-Agnostic Assessment (Content check).
3.  **Phase 4 (Moved)**: Narrative & Pacing Gaps.
    *   *Effect*: Each path is fully fleshed out with all necessary intermediate beats (`Gap 1`, `Gap 2`, etc.). The causal chain for each timeline is complete and robust.
4.  **Phase 3 (Moved)**: Intersection Detection.
    *   *Effect*: The algorithm now compares two *complete* chains.
    *   *Benefit*: We can enforce a strict **Compatibility Check**. If `Beat X` (Path A) requires `Gap 1`, and `Beat Y` (Path B) does not, they **cannot be merged** unless `Gap 1` is also compatible with Path B.
    *   *Result*: The "Conditional Requires" paradox becomes impossible. Incompatible merges are simply rejected, keeping the timelines distinct where they differ structurally.

**Implementation Impact:**
- **Pipeline Config**: Update `GrowStage._phase_order()` in `src/questfoundry/pipeline/stages/grow.py`.
- **Algorithms**: Ensure `build_intersection_candidates` works correctly on graphs that already contain gap beats (should be supported as they are just standard beats).
- **Validation**: The existing `check_intersection_compatibility` logic will now naturally catch these cases (because the `requires` edges from gaps will already exist).

This aligns the pipeline with the principle of "Local Coherence before Global Integration."

### LINKED ISSUE #359: Design: GROW invariant for intersections (no conditional prerequisites for shared beats)
### Why this issue exists
Discussion in #357/#358 keeps drifting toward patching symptoms (arc enumeration tweaks, successor filtering, etc.). This issue tracks a *deeper* procedure-level framing to prevent a whack-a-mole cycle.

### Core failure mode (class)
A beat becomes **shared** (multi-path) during GROW, but later steps introduce or expose `requires` prerequisites that are **not shared** across the same set of paths. That creates “conditional prerequisites” which then get dropped in some arcs/subsets, causing inconsistent ordering across arcs and downstream failures (e.g. `passage_dag_cycles`).

### Proposed root-cause framing
Treat intersection/weaving as an operation that must preserve a single structural invariant:

**Invariant: No conditional prerequisites**
For any `requires` edge `A -> B`:
- `B` must be at least as “shared” as `A`.
- Put differently: a beat that is shared across a set of paths may not require a beat that exists only on a strict subset of those paths.

This makes “a shared scene depends on path-only lead-in” an invalid state by construction.

### Procedure-level proposal (GROW)
1) **Complete each path first** (aligns with the spirit of #358):
   - Run scene-type tagging + narrative/pacing gaps to make each path’s chain coherent before any weaving.

2) **Intersections become alignment/merge of completed chains**:
   - Only allow an intersection/merge if the prerequisite neighborhood of the shared scene is compatible across all involved paths.

3) **When incompatibility appears, only structural moves are allowed** (no ad-hoc `requires` edges that create conditionality):
   - Keep beats separate (no intersection), OR
   - Lift prerequisites into the shared set (merge prerequisite beats too), OR
   - Split into `path-specific lead-in` beats feeding a truly shared beat (shared beat no longer depends on path-only nodes).

### How this differs from “symptom fixes”
- It doesn’t rely on arc enumeration filtering/dedup heuristics.
- It defines what intersections *mean* relative to ordering, and prevents producing unstable graphs in the first place.

### Acceptance criteria (behavioral)
- GROW never produces a graph where a multi-path beat has a prerequisite beat that is missing from some of its paths.
- If an intersection would violate the invariant, GROW must choose one of the structural alternatives (separate / lift / split) instead.

### Related
- #357: observed `passage_dag_cycles` failure mode.
- #358: phase reordering proposal (gaps before intersections) is compatible, but not sufficient unless this invariant is enforced.

pvliesdonk:
## Problem

GROW validation fails with `passage_dag_cycles` because shared (intersection) beats can acquire `requires` edges to path-specific gap beats. When arc enumeration builds beat sets for arcs that don't include the gap beat's path, the `requires` edge is silently dropped, producing contradictory successor orderings across arcs and cycles in the passage DAG. Additionally, seed pruning was not updating `considered`/`implicit` fields on demoted dilemmas, causing a validation mismatch.

## Changes

**Commit 1** (existing): Update `considered`/`implicit` when pruning demoted dilemmas

**Commit 2** (new):
- Reorder `_phase_order()` in `grow.py`: gap detection (4a/4b/4c) now runs before intersection detection (3), so each path is fully elaborated before cross-path weaving
- Add no-conditional-prerequisites invariant to `check_intersection_compatibility()`: rejects intersections where a shared beat would depend on a path-specific beat (`paths(prerequisite) ⊇ paths(dependent_after_merge)`)
- Add `make_conditional_prerequisite_graph()` test fixture
- Add `TestConditionalPrerequisiteInvariant` with 6 regression tests
- Update `docs/design/procedures/grow.md` with execution order note, invariant spec, and references to future alternatives (#360, #361)

## Not Included / Future PRs

- **Lift prerequisites into shared set** (#360): merge prerequisite beat into all intersection paths
- **Split path-specific lead-ins** (#361): create path-specific copies of the shared beat
- Both are referenced in code comments and design docs

## Test Plan

```bash
uv run mypy src/questfoundry/  # ✅ clean
uv run ruff check src/         # ✅ clean
uv run pytest tests/unit/test_grow_algorithms.py -x -q  # ✅ 130 passed
uv run pytest tests/unit/test_grow_validation.py -x -q  # ✅ 33 passed
pre-commit run --files ...     # ✅ all passed
```

New tests:
- `test_rejects_intersection_with_conditional_prerequisite`
- `test_accepts_intersection_without_conditional_prerequisites`
- `test_accepts_when_prerequisite_spans_all_paths`
- `test_phase_order_gaps_before_intersections`
- `test_rejects_when_prerequisite_has_no_paths`
- `test_reports_multiple_conditional_prerequisites`

## Risk / Rollback

- Phase reordering is safe: architect review confirmed no data dependencies from phases 4a/4b/4c on phase 3 output
- The invariant check is additive (new rejection, no existing behavior changed)
- Rollback: revert both commits; issues #357/#358/#359 would need reopening

Closes #357, closes #358, closes #359

### DIFF
--- a/docs/design/procedures/grow.md
+++ b/docs/design/procedures/grow.md
@@ -109,6 +109,14 @@ If the non-canonical answer is not promoted to a path in SEED, that dilemma has 
 
 ## Algorithm Phases
 
+> **Execution order note:** Phases 4a/4b/4c (gap detection & scene-type
+> tagging) execute **before** Phase 3 (intersection detection).  This
+> ensures each path is fully elaborated with gap beats before cross-path
+> weaving, preventing "conditional prerequisites" where a shared beat
+> depends on a path-specific gap beat.  Phase numbering is preserved for
+> historical continuity; execution order is defined in `_phase_order()`.
+> See also: **No-Conditional-Prerequisites Invariant** under Phase 3.
+
 ### Phase 1: Beat Graph Import
 
 **Input:** SEED artifacts
@@ -196,6 +204,7 @@ A beat between `reveals` and `commits` may be **logically agnostic** (player has
      - No `requires` conflicts (A requires B, B requires A)
      - No timing contradictions
      - Location resolution possible (shared location exists in both location sets)
+     - **No conditional prerequisites** (see invariant below)
    - If compatible: propose as intersection with resolved location
 4. Human reviews proposed intersections:
    - Approve: execute intersection operation (mark or merge), set resolved location
@@ -217,13 +226,36 @@ A beat between `reveals` and `commits` may be **logically agnostic** (player has
 
 **Iteration:** One pass of clustering. If human wants more intersections, re-run with different guidance.
 
+#### No-Conditional-Prerequisites Invariant
+
+For any `requires` edge A → B where A is in a proposed intersection:
+
+> **`paths(B) ⊇ paths(A_post_intersection)`**
+
+A beat that would become shared across multiple paths (via intersection
+marking) cannot depend on a beat that exists only on a strict subset of
+those paths.  If this invariant is violated, the `requires` edge would be
+silently dropped during arc enumeration for arcs missing the prerequisite's
+path.  This produces inconsistent topological orderings across arcs and
+causes `passage_dag_cycles` failures in validation.
+
+**Current strategy:** Reject the intersection.  The beats remain separate.
+
+**Future alternatives** (not yet implemented):
+- **Lift prerequisites (#360):** Also add the prerequisite beat to all
+  intersection paths, making it globally shared.  Risk: may widen a beat
+  beyond its narrative intent.
+- **Split lead-ins (#361):** Create path-specific copies of the shared beat
+  so the path-specific version keeps the dependency while others are
+  independent.  Risk: increases beat count and complexity.
+
 ---
 
 ### Phase 4: Gap Detection and Scene-Type Tagging
 
 **Purpose:** Find missing beats needed for narrative continuity AND assign scene types for pacing.
 
-**Input:** Beat graph with intersections
+**Input:** Beat graph (before intersections — see execution order note above)
 
 **Operations:**
 
--- a/src/questfoundry/graph/grow_algorithms.py
+++ b/src/questfoundry/graph/grow_algorithms.py
@@ -801,23 +801,71 @@ def check_intersection_compatibility(
             )
         )
 
-    # Check no circular requires between the intersection beats
+    # Check requires edges originating from intersection beats.
+    # Two checks in one pass: (1) no circular requires between intersection
+    # beats, and (2) no conditional prerequisites (see below).
     beat_set = set(beat_ids)
-    requires_edges = graph.get_edges(from_id=None, to_id=None, edge_type="requires")
-    for edge in requires_edges:
-        from_id = edge["from"]
-        to_id = edge["to"]
-        if from_id in beat_set and to_id in beat_set:
-            errors.append(
-                GrowValidationError(
-                    field_path="intersection.requires",
-                    issue=(
-                        f"Beat '{from_id}' requires '{to_id}' — "
-                        f"intersection beats cannot have requires dependencies on each other"
-                    ),
-                    category=GrowErrorCategory.STRUCTURAL,
+
+    # Build beat→paths mapping for the conditional-prerequisites check.
+    belongs_to_edges = graph.get_edges(from_id=None, to_id=None, edge_type="belongs_to")
+    beat_paths: dict[str, set[str]] = {}
+    for edge in belongs_to_edges:
+        beat_paths.setdefault(edge["from"], set()).add(edge["to"])
+
+    # The union of all paths that the intersection would span
+    union_paths: set[str] = set()
+    for bid in beat_ids:
+        union_paths.update(beat_paths.get(bid, set()))
+
+    # Iterate only over outgoing requires edges from intersection beats
+    # (targeted lookups instead of scanning all requires edges).
+    for from_id in beat_set:
+        for edge in graph.get_edges(from_id=from_id, to_id=None, edge_type="requires"):
+            to_id = edge["to"]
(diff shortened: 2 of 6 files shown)


pvliesdonk (commented):
Addressed Gemini's performance suggestion: refactored both the circular-requires check and conditional-prerequisites check into a single pass over outgoing `requires` edges from `beat_set` members only, eliminating the full-graph scan of all requires edges. Also fixed the pre-existing `test_phase_order_names` test that expected the old phase order.

Issue #366, opened 2026-01-29
feat(prompts): strengthen brainstorm/seed guidance for entity interconnection and convergence


pvliesdonk:
## Problem

Analysis of 5 test runs shows that story structure quality (intersections, spine arc, branching) is determined by how well the brainstorm and seed stages create mechanically interconnected entities. test-5 succeeded because entities had position-based contradictions tied to specific locations; test-3/4 had thematically rich but structurally loose entities.

Current prompts encourage thematic diversity but not mechanical interconnection:
- `discuss_brainstorm.yaml:19-22`: "aim for 4-6 distinct settings" — focuses on quantity, not connectivity
- `discuss_seed.yaml:90`: "Hint at where paths should merge" — too weak, test-3/4 ignored it

## What test-5 did right (the target pattern)

1. **Characters have position-based contradictions**: Dr. Voss → Bio-Secure Lab (investigation), Kaito → Maintenance Tunnels (hidden recorder)
2. **Locations serve dual purposes**: Bio-Secure Lab = crime scene AND fire origin, Atrium = public gathering AND clock mystery
3. **Binary factional opposition** forces convergence: Silence Collective vs. Open Data Movement
4. **Explicit forced path**: Atrium → Bio-Secure Lab → Tunnels → Herbarium

### Brainstorm prompt (`discuss_brainstorm.yaml`)
Add interconnection guidance:
- Characters from different dilemmas should have reasons to visit the same locations
- Key locations should serve at least 2 dilemmas
- Give characters position-based leverage (access, possession, authority)
- Avoid characters who can investigate independently without encountering others

**Prompt-only change.** No validation needed — GROW handles convergence at runtime. The guidance shapes LLM thinking upstream.

### Seed discuss prompt (`discuss_seed.yaml`)
Strengthen convergence requirement:
- Replace "Hint at where paths should merge" with explicit requirement for location-based convergence points
- Require 2-4 locations where characters from different dilemmas MUST appear
- Map a forced investigation path: start → evidence → confrontation

**Prompt-only change initially.** If LLMs consistently ignore this guidance, follow up with validation in `mutations.py` to check convergence_sketch references valid locations.

### Seed serialize convergence section (`serialize_seed_sections.yaml`)
Add explicit good/bad examples:
```
GOOD: "By Act 2, paths converge at archive_vault when all answers lead to final choice"
BAD: "Paths merge emotionally" (abstract, no location)
[...]

Issue #365, opened 2026-01-29
feat(grow): add quality gates for spine arc and intersection acceptance


pvliesdonk:
## Problem

The GROW stage currently always completes successfully even when critical narrative structure is missing. In test runs:

- **test-3 and test-4**: No spine arc found, no divergence/convergence points — pipeline reported "validation passed with warnings"
- **test-4**: All 10 proposed intersections rejected (0% acceptance) — pipeline still completed with `status="completed"`
- Prune phase falls back to "Cannot BFS without spine; all passages kept" — effectively disabling reachability analysis

The result is a "complete" story that's actually a collection of parallel linear paths, not a branching narrative.

## Evidence

- test-4 debug log: 9x `phase3_incompatible_intersection` warnings, 0 applied, stage still "completed"
- test-3/4: `spine_arc_id=None` in GrowResult, no validation check for this
- test-3/4: 0 divergence/convergence points vs test-5's 15 each
- Current validation (`grow.py:1771-1796`) only checks structural reachability, not narrative quality

### Tier 1: Fail on missing foundations
1. **Spine arc check** after `enumerate_arcs` (phase 5): If no arc has `arc_type == "spine"`, fail the phase with actionable error
2. **Intersection rejection check** after phase 3: If `proposed > 0 and applied == 0`, fail with detail about why all proposals were rejected

### Tier 2: Warn on degraded quality  
3. Add validation check for divergence/convergence points (warn if 0 found)
4. Add `spine_arc_exists` to the validation phase checks

### Implementation notes
- Phase 3 already tracks `applied_count` and `skipped_count` — just needs a conditional on the return
- Phase 5 already identifies spine arcs — just needs a check after enumeration
- Validation phase (`_phase_10_validation`) needs new check entries

## Complexity
Small — conditional checks at existing decision points, no architectural changes.

Issue #364, opened 2026-01-29
fix(seed): inject valid answer IDs per dilemma before paths serialization


pvliesdonk:
## Problem

When serializing the paths section during SEED, the LLM must match each path's `answer_id` to the parent dilemma's `considered` array. However, no structured manifest of valid answer IDs is injected into the serialization context — the LLM must infer valid answers from the prose summary alone.

This caused test-2 to fail with 4 persistent path-answer mismatches across all 3 retry loops (e.g., `deceptive` proposed for a dilemma with only `['honest']` in considered). The LLM never corrected the errors because it couldn't see the constraint clearly.

## Evidence

- test-2 failure: paths.1.answer_id `deceptive` not in dilemma considered list `['honest']` (4 similar errors)
- 3 outer retries exhausted without improvement — feedback wasn't actionable enough
- The prompt at `serialize_seed_sections.yaml:120-125` documents the invariant but doesn't inject the data

## Proposed Fix

1. Add `format_answer_ids_by_dilemma()` in `graph/context.py` that builds a manifest from the serialized dilemmas section:
   ```
   ### Valid Answer IDs per Dilemma
   - dilemma::voss_honest_or_deceptive → considered: [honest], implicit: [deceptive]
   ```
2. In `agents/serialize.py`, after dilemmas are serialized (~line 1185), inject this manifest into the context used for paths serialization
3. No prompt template changes needed — the constraint documentation already exists

Related to #363 (SEED validation timing). This improvement prevents the error; #363 ensures errors that still occur get caught during serialize rather than post-write.

## Complexity
Small — one new formatting function + one injection point in the serialize loop.

Issue #363, opened 2026-01-29
fix(seed): move cross-reference validation into serialize phase's repair loop


pvliesdonk:
## Problem

SEED stage semantic validation (cross-reference checks like entity ID resolution and path-dilemma answer matching) runs **after** the artifact is written to disk, in the orchestrator — not during the serialize phase where the LLM could correct errors. This means the serialize phase's 3-retry validation loop never sees these errors, and the LLM has no opportunity to fix them.

## Evidence

**test-1**: `entity::vault` used instead of `entity::loc_vault` — caught only after artifact write. The serialize phase accepted the output as valid, and the cross-reference error was detected too late.

**test-2**: Path answers not in dilemma `considered` lists (4 errors) — persisted across all 3 retries because the serialize phase never surfaced them to the LLM.

**test-3/4/5**: Succeeded, but only because their errors happened to be caught within the serialize phase by coincidence.

Debug log explicitly confirms the gap:
```
"seed_validation_bypassed", "msg": "SeedMutationError reached orchestrator - validation should happen during serialize"
```

## Impact

- test-1 and test-2 both **FAILED** despite generating otherwise usable output
- ~30 minutes and ~90K tokens wasted per failed run (retry loops that could never succeed)
- False negatives undermine confidence in the pipeline

## Proposed Fix

Move cross-reference validation (entity ID resolution, path-dilemma answer matching) into the serialize phase's validation loop so that:

1. Validation errors are surfaced to the LLM as structured feedback
2. The LLM gets a chance to correct errors within the existing retry budget
3. Only fully validated artifacts are written to disk

This aligns with the existing pattern where the serialize phase runs `_validate_and_repair()` with structured error details that guide LLM correction.

## Not Included / Future Work

- Refactoring other stages to follow the same pattern (audit separately)
- Adding new cross-reference checks beyond existing ones
- Changing the retry budget (3 retries is sufficient if errors are surfaced correctly)

PR #367, opened 2026-01-29
feat: SEED/GROW quality improvements (#360, #361, #363, #364, #365, #366)

pvliesdonk:
## Problem

Test project analysis (test-1 through test-5) revealed systematic quality issues in SEED and GROW stages:
- SEED cross-reference errors (path answer_id not in dilemma considered) escape the serialize retry loop because only the downstream section is retried, not the upstream cause
- No answer ID manifest injected before paths serialization, so models invent phantom answer_ids
- Brainstorm/seed prompts lack guidance on cross-dilemma interconnection, producing isolated storylines
- GROW completes with status "completed" even when all intersections are rejected or no spine arc exists
- Intersection compatibility check unconditionally rejects conditional prerequisites instead of attempting recovery

## Changes

Six stacked commits, each addressing one or two issues:

1. **Prompt improvements (#366)**: Add interconnection guidance to brainstorm prompt, explicit location-based convergence to seed prompt, GOOD/BAD examples to convergence serialization
2. **Answer ID manifest (#364)**: `format_answer_ids_by_dilemma()` injected after dilemmas section, before paths serialization
3. **Cross-reference validation fix (#363)**: Propagate check-11c errors to upstream dilemmas section, sort retries by dependency order, refresh context between retries
4. **GROW quality gates (#365)**: Fail phase 3 on 100% intersection rejection, fail phase 5 on missing spine arc, add `check_spine_arc_exists()` validation
5. **Lift/split prerequisites (#360, #361)**: `_try_lift_prerequisite()` widens prereqs to all paths (transitive, max depth 3); `_try_split_beat()` creates path-specific variants; fallback chain: lift → split → reject
6. **Test fixups**: Updated test expectations for new quality gate behavior

## Not Included / Future PRs

- No changes to GROW LLM prompts (intersection failures are structural, not prompt-related)
- No changes to the outer seed retry loop or orchestrator error handling
- No end-to-end validation against test projects (requires manual `qf run --to grow`)

## Test Plan

- `uv run mypy src/` — clean (63 files)
- `uv run ruff check src/` — clean
- `uv run pytest tests/unit/ -x -q` — 1345 passed
- New tests: 5 for `format_answer_ids_by_dilemma`, 3 for `check_spine_arc_exists`, 8 for lift/split/transitive strategies

## Risk / Rollback

- Lift/split mutates the graph during compatibility checking — this is intentional (the mutation IS the recovery), but means `check_intersection_compatibility` is no longer side-effect-free
- Quality gates change GROW from "always completes" to "fails on structural deficiencies" — pipelines that previously produced degraded output will now stop earlier
- Cross-reference error propagation adds synthetic dilemma errors during retry, which could cause unnecessary dilemma re-serialization if the model's considered list was correct

## Review Guide

Suggested review order: commits 1→2→4→3→5→6 (prompts, then simple code, then complex code)

### DIFF
--- a/prompts/templates/discuss_brainstorm.yaml
+++ b/prompts/templates/discuss_brainstorm.yaml
@@ -21,6 +21,17 @@ system: |
   - If something has a physical space where scenes can occur, consider making it a location not an object
   - A story with only 1-2 locations feels confined; more settings enable natural scene variation and convergence points
 
+  ## Interconnection Guidance
+  Dilemmas should NOT be isolated storylines. Build interconnection into the material:
+  - **Shared locations across dilemmas**: At least 2-3 locations should be relevant to multiple dilemmas
+    (e.g., an archive is where both the trust investigation AND the artifact mystery converge)
+  - **Position-based contradictions**: Characters who appear trustworthy under one dilemma
+    but suspicious under another create natural cross-dilemma tension
+  - **Dual-purpose locations**: A location that serves one function for one dilemma and a
+    different function for another (e.g., a garden is a meeting place AND a hiding spot)
+  - **Avoid isolated investigators**: If a character is central to only one dilemma with no
+    connection to others, the story will fragment into parallel tracks that never intersect
+
   2. **Dilemmas** - Binary dramatic questions that drive the story
      - Each dilemma is a yes/no question (e.g., "Can the mentor be trusted?")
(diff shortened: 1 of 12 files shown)


pvliesdonk:
Splitting into separate PRs per commit for independent review.

PR #371, opened 2026-01-29
feat(grow): add quality gates for spine arc and intersections (#365)

pvliesdonk:
## Problem
GROW silently succeeds even when all proposed intersections are rejected (no cross-dilemma overlap) or when no spine arc is created (no arc covers all canonical paths). These are structural failures that should prevent downstream stages from running on invalid story graphs.

## Changes
- `grow.py` (`_phase_3_intersections`): Fail when proposals exist but all are rejected — "All N proposed intersections rejected"
- `grow.py` (`_phase_5_enumerate_arcs`): Fail when arcs exist but none is a spine arc — "No spine arc created"
- `grow_validation.py`: Add `check_spine_arc_exists()` validation check, warns when no arcs exist (empty graph), fails when arcs exist but none is spine

## Not Included / Future PRs
- Intersection recovery strategies (#360, #361)
- Minimum intersection acceptance rate threshold

## Test Plan
- `uv run pytest tests/unit/test_grow_validation.py -x -q` — 3 new spine arc tests
- `uv run mypy src/ && uv run ruff check src/`

## Risk / Rollback
- Medium risk: previously-passing GROW runs may now fail if they lack intersections or spine arcs
- This is intentional — these are structural defects that should be caught early
- Rollback: remove the two quality gate checks

Closes #365 (partially — test fixups in #373)

> **Stack**: 4/6 — [#368](#368) → [#369](#369) → [#370](#370) → this PR → [#372](#372) → [#373](#373)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #370, opened 2026-01-29
fix(seed): propagate cross-reference errors to upstream sections (#363)

pvliesdonk:
## Problem
Cross-reference errors between sections (e.g., a path's `answer_id` not in its dilemma's `considered` list) are detected during serialize but only retry the downstream section (paths). The upstream section (dilemmas) is never retried, so the error persists and escapes to the orchestrator as `SeedMutationError`.

## Changes
- `serialize.py`: Add `_SECTION_ORDER` constant defining dependency order
- `serialize.py`: Add `_propagate_cross_section_errors()` — detects check-11c errors in paths section and creates synthetic dilemma errors to trigger upstream retry
- `serialize.py`: Modified `_group_errors_by_section()` to call propagation and sort retry sections by dependency order
- `serialize.py`: Added dilemma context refresh logic on retry (re-inject answer IDs after dilemma section is re-serialized)

## Not Included / Future PRs
- Additional cross-reference checks beyond check-11c
- GROW quality gates (#365)

## Test Plan
- `uv run mypy src/ && uv run ruff check src/`
- Behavioral testing requires LLM integration; unit validation covered by existing mutation tests

## Risk / Rollback
- Medium risk: changes error retry ordering in the serialize loop
- Rollback: remove `_propagate_cross_section_errors()` and revert `_group_errors_by_section()` changes

Closes #363

> **Stack**: 3/6 — [#368](#368) → [#369](#369) → this PR → [#371](#371) → [#372](#372) → [#373](#373)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #369, opened 2026-01-29
feat(seed): inject valid answer IDs before paths serialization (#364)

pvliesdonk:
## Problem
SEED serialization produces "phantom ID" errors where the LLM invents answer IDs not present in any dilemma. The serialize loop injects valid entity IDs and path IDs before their respective sections, but answer IDs per dilemma are not provided, leaving the LLM to guess.

## Changes
- `context.py`: Add `format_answer_ids_by_dilemma()` — builds a manifest of valid answer IDs (considered + implicit) per dilemma
- `serialize.py`: Inject the answer ID manifest into `enhanced_brief` after dilemmas section is serialized, before paths section

## Not Included / Future PRs
- Cross-reference validation timing (#363)
- GROW quality gates (#365)

## Test Plan
- `uv run pytest tests/unit/test_graph_context.py -x -q` — 5 new tests for `format_answer_ids_by_dilemma`
- `uv run mypy src/ && uv run ruff check src/`

## Risk / Rollback
- Low risk: additive change, injects additional context into the serialize prompt
- Rollback: remove the injection block in serialize.py

Closes #364

> **Stack**: 2/6 — [#368](#368) → this PR → [#370](#370) → [#371](#371) → [#372](#372) → [#373](#373)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #368, opened 2026-01-29
feat(prompts): strengthen brainstorm/seed prompts for interconnection (#366)

pvliesdonk:
## Problem
Test project analysis (test-3, test-4) showed SEED produces stories with isolated dilemmas and abstract convergence statements instead of concrete location-based intersections. The brainstorm and seed prompts lack explicit guidance for interconnection.

## Changes
- `discuss_brainstorm.yaml`: Add "Interconnection Guidance" section requiring shared locations across dilemmas, position-based contradictions, dual-purpose locations
- `discuss_seed.yaml`: Replace weak "Hint at where paths should merge" with explicit convergence requirements (2-4 location-based convergence points, forced investigation chain)
- `serialize_seed_sections.yaml`: Enhance convergence_prompt with GOOD/BAD examples requiring location-based convergence points

## Not Included / Future PRs
- Code changes to enforce convergence (see #363, #364, #365)
- GROW-side intersection improvements (see #360, #361)

## Test Plan
- `uv run mypy src/ && uv run ruff check src/` (prompts don't have unit tests)
- Prompt-only changes; no behavioral code modified

## Risk / Rollback
- Low risk: prompt-only changes, no code behavior affected
- Rollback: revert prompt files

Closes #366

> **Stack**: 1/6 — base: `main` → [#369](#369) → [#370](#370) → [#371](#371) → [#372](#372) → [#373](#373)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #372, opened 2026-01-29
feat(grow): add lift/split recovery for conditional prerequisites (#360, #361)

### LINKED ISSUE #360: Enhancement: Lift prerequisites into shared set during intersection

### LINKED ISSUE #361: Enhancement: Split path-specific lead-ins for conditional prerequisites

pvliesdonk:
## Problem
`check_intersection_compatibility()` unconditionally rejects intersections where a shared beat has a prerequisite that doesn't span all intersection paths. This is overly strict — many conditional prerequisites can be resolved by widening the prerequisite's path membership (lift) or splitting the beat into variants (split).

## Changes
- `grow_algorithms.py`: Add `_try_lift_prerequisite()` — recursively widens prerequisite `belongs_to` edges to cover all intersection paths (max depth 3, cycle-safe)
- `grow_algorithms.py`: Add `_try_split_beat()` — creates a `{beat_id}_split` variant node for narrow-path cases when lift fails
- `grow_algorithms.py`: Modified `check_intersection_compatibility()` to try lift → split → reject (sequential fallback)

## Not Included / Future PRs
- FILL stage handling of split beats (the `split_from` field is set for downstream use)
- Cycle detection beyond simple path traversal

## Test Plan
- `uv run pytest tests/unit/test_grow_algorithms.py::TestConditionalPrerequisiteInvariant -x -q` — 5 tests updated/added:
  - `test_lifts_conditional_prerequisite` (simple lift)
  - `test_lifts_orphan_prerequisite` (lift without existing paths)
  - `test_lifts_multiple_conditional_prerequisites` (multiple prereqs)
  - `test_lift_transitive_prerequisite` (chain lift)
  - `test_rejects_when_lift_depth_exceeded` (depth > 3 rejected)
- `uv run mypy src/ && uv run ruff check src/`

## Risk / Rollback
- Medium risk: changes intersection acceptance behavior (more permissive)
- Graph mutations (new edges, new nodes) are additive — won't break existing structures
- Rollback: revert `check_intersection_compatibility()` changes

Closes #360, closes #361

> **Stack**: 5/6 — [#368](#368) → [#369](#369) → [#370](#370) → [#371](#371) → this PR → [#373](#373)

### DIFF
--- a/src/questfoundry/graph/grow_algorithms.py
+++ b/src/questfoundry/graph/grow_algorithms.py
@@ -22,6 +22,9 @@ from typing import TYPE_CHECKING, Any
 from questfoundry.graph.context import normalize_scoped_id
 from questfoundry.graph.mutations import GrowErrorCategory, GrowValidationError
 from questfoundry.models.grow import Arc
+from questfoundry.observability.logging import get_logger
+
+log = get_logger(__name__)
 
 if TYPE_CHECKING:
     from questfoundry.graph.graph import Graph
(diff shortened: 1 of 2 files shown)


pvliesdonk:
Closing stacked PR — consolidating into a single PR with review feedback addressed. The stacked PR approach via git-branchless didn't work reliably (see CLAUDE.md update in ea62328).

PR #413, opened 2026-01-30
feat(size): add size profiles + CLI improvements

pvliesdonk:
## Problem
Story generation needs coordinated size control across all pipeline stages, and the CLI
lacks convenience features for project initialization and batch execution.

### Size Profile System (core feature)
- Add `SizeProfile` dataclass with 4 presets: vignette, short, standard, long
- Add `story_size` field to DREAM `Scope` model
- Parameterize BRAINSTORM, SEED, GROW, and FILL templates with size-aware ranges
- Apply size limits to code-side algorithms (grow beat caps, fill vision context)
- Fix FILL vision context bug (was passing raw graph node instead of formatted string)

### CLI Improvements (3 commits)
- **DREAM preset guidance**: Add size preset descriptions to discuss prompt so the LLM
  can guide users toward the right story size during DREAM
- **`qf init --provider`**: Add `--provider` flag, remove redundant `pipeline.stages`
  from project.yaml, extract `_init_project()` helper
- **`qf run --init`**: Auto-create projects with `--init` flag, interactive Y/n prompt
  in TTY mode, or hard fail in non-interactive mode

## Not Included / Future PRs
- DREAM-specific serialize prompt with explicit preset validation
- "epic" preset for large models
- Per-project size range overrides in project.yaml

## Test Plan
- `uv run pytest tests/unit/test_size.py -x -q` — 259 lines of size profile tests (all pass)
- `uv run pytest tests/unit/test_cli.py -x -q` — 71 tests including new --provider, --init tests (all pass)
- `uv run pytest tests/unit/test_discuss_agent.py -x -q` — discuss prompt tests (all pass)
- `uv run mypy src/questfoundry/` — clean
- `uv run ruff check src/` — clean
- All pre-commit hooks pass on every commit

## Risk / Rollback
- **Backward compatible**: Existing projects with `pipeline.stages` in YAML still work
  (field is parsed but ignored; config defaults to `DEFAULT_STAGES`)
- Size profiles use template variables — if a template doesn't reference them, they're
  silently unused (no runtime errors)
- `--init` flag is opt-in; default behavior unchanged for `qf run`

## Review Guide
Suggested review order:
1. `src/questfoundry/pipeline/size.py` + `tests/unit/test_size.py` — core data model
2. `src/questfoundry/models/dream.py` — `story_size` field addition
3. `prompts/templates/*.yaml` — template parameterization
4. `src/questfoundry/agents/prompts.py` — size var injection
5. `src/questfoundry/cli.py` + `tests/unit/test_cli.py` — CLI improvements

### DIFF
--- a/prompts/templates/discuss.yaml
+++ b/prompts/templates/discuss.yaml
@@ -10,9 +10,9 @@ system: |
   - Genre and tone
   - Themes and motifs
   - Target audience
-  - Scope and complexity
+  - Scope and complexity (story size preset)
   - Style notes
-
+  {size_presets_section}
   ## Guidelines
   - Ask clarifying questions to understand the vision
   - Suggest possibilities but respect the user's preferences
@@ -90,5 +90,16 @@ research_tools_section: |
     ]
   )
   ```
+
+size_presets_section: |
+  ## Story Size Presets
+  When discussing scope, guide the user toward one of these size presets:
+  - **vignette**: 5-15 passages, 2000-5000 words. Tight single-scene exploration with minimal branching.
+  - **short**: 15-30 passages, 5000-15000 words. Focused story with limited branching paths.
+  - **standard** (default): 30-60 passages, 15000-30000 words. Full branching narrative with multiple paths.
+  - **long**: 60-120 passages, 30000-60000 words. Expansive world with deep branching and many characters.
+
+  The chosen preset determines entity counts, word targets, and branching complexity
+  for all downstream stages. If the user doesn't express a preference, default to "standard".
 
 components: []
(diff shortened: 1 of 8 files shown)


pvliesdonk:
Addressed review feedback in commit 0a2b525:

1. **story_size Literal validation** (claude) — Applied. Now uses `Literal["vignette", "short", "standard", "long"]` to catch invalid LLM output at validation time.
2. **Rename dream_data → vision_data** (gemini) — Applied.
3. **Simplify resolve_size_from_graph** (gemini) — Applied with `or {}` pattern.
4. **Hardcoded max_arcs=16 fallback** (claude) — Applied. Now uses `get_size_profile("standard").max_arcs`.
5. **Explicit size_profile parameter** (claude) — Deferred to a follow-up PR. This touches all 4 stage `execute()` signatures for a non-functional type-safety improvement. The `kwargs` pattern works correctly at runtime.

PR #453, opened 2026-01-31
fix(grow): add explicit dilemma tags to intersection beat summaries

pvliesdonk:
## Problem
GROW Phase 3 (intersections) fails when the LLM groups beats by dilemma instead of cross-dilemma. The beat summaries sent to the LLM include beat ID, summary, location, and entities — but not which dilemma the beat belongs to.

The dilemma is only inferrable from the beat ID naming convention (`beat::dilemma_name__answer_beat_NN`), which qwen3:4b parses unreliably. The model groups beats sharing a name prefix (e.g. all `black_archive_*` beats) rather than mixing beats from different dilemmas as required.

**Evidence**: `test-20260131T1159` has 52 beats pre-intersection, 4 locations with cross-dilemma overlap, but all 10 LLM proposals were single-dilemma — rejected by the validator.

## Changes
- Extract primary dilemma ID from each beat's `dilemma_impacts` and include as `[dilemma: ...]` tag in the beat summary line
- Add "Dilemma Matching" section to intersection prompt with explicit good/bad examples
- Rephrase "What NOT to Do" to reference the dilemma tags directly

## Not Included / Future PRs
- No changes to the validator logic (already correctly rejects same-dilemma intersections)
- No changes to `build_intersection_candidates` algorithm

## Test Plan
- `uv run mypy src/questfoundry/pipeline/stages/grow.py` — passes
- `uv run pytest tests/unit/test_grow*.py -x -q` — 362 passed

## Risk / Rollback
- Low risk — only changes context sent to LLM, no logic changes
- Validator remains the safety net for invalid proposals

### DIFF
--- a/prompts/templates/grow_phase3_intersections.yaml
+++ b/prompts/templates/grow_phase3_intersections.yaml
@@ -19,14 +19,22 @@ system: |
   5. CRITICAL: Each beat can appear in at most ONE intersection. Never assign the
      same beat to multiple intersections.
 
+  ## Dilemma Matching (CRITICAL)
+  Each beat below is tagged with `[dilemma: ...]`. Use this tag to ensure
+  every intersection mixes beats from DIFFERENT dilemmas:
+
+  GOOD: beat from [dilemma: dilemma::A] + beat from [dilemma: dilemma::B] = valid (2 different dilemmas)
+  BAD:  beat from [dilemma: dilemma::A] + beat from [dilemma: dilemma::A] = INVALID (same dilemma)
+
   ## What NOT to Do
-  - Do NOT group beats from the same dilemma (those are paths, not intersections)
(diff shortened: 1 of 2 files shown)


pvliesdonk:
Addressed review feedback in 26f9b5d:

- **Gemini**: Adopted sorted set comprehension for deterministic dilemma tag selection and deduplication.
- **Claude**: The `dilemma_ids[0] if dilemma_ids else "unknown"` pattern already handles the empty `dilemma_impacts` case safely — beats with no impacts get tagged `"unknown"`. These beats are correctly filtered out by the existing cross-dilemma validator downstream, so tagging them `"unknown"` is the right behavior (they can't form valid intersections). The dict key check (`"dilemma_id" in impact`) works correctly here since graph data is dict-based, not Pydantic models.

PR #454, opened 2026-01-31
fix(dream): make scope required so DREAM always sets story_size

pvliesdonk:
## Problem
When a user requests a specific story size (e.g. "vignette-length"), the DREAM discuss and summarize phases capture it in prose ("Vignette-Length Pulp Space Opera"), but the serialize phase omits the `scope` field because:

1. `scope: Scope | None = None` in `DreamArtifact` — optional, so LLM skips it
2. The serialize prompt mentions "scope" but never shows the Scope object structure
3. `resolve_size_from_graph()` silently falls back to "standard"

**Result**: User asks for vignette → all downstream stages use standard size ranges.

## Changes
- Give all `Scope` fields sensible defaults (standard preset values) so `Scope()` is valid without arguments
- Change `DreamArtifact.scope` from `Scope | None = None` to `Scope = Field(default_factory=Scope)`
- Add explicit Scope schema example to serialize prompt with size-to-values mapping
- Update tests to reflect scope always being present with default values

## Not Included / Future PRs
- No changes to `resolve_size_from_graph()` (already reads `vision.scope.story_size` correctly)
- No changes to discuss/summarize prompts (already mention size presets)

## Test Plan
- `uv run mypy src/questfoundry/models/dream.py` — passes
- `uv run pytest tests/unit/test_artifacts.py tests/unit/test_dream_stage.py tests/unit/test_mutations.py tests/unit/test_size.py -x -q` — 218 passed

## Risk / Rollback
- Low risk — Scope defaults match the previous fallback behavior ("standard")
- Existing projects without scope in graph.json are unaffected (`resolve_size_from_graph` still falls back gracefully)

### DIFF
--- a/prompts/templates/serialize.yaml
+++ b/prompts/templates/serialize.yaml
@@ -29,9 +29,10 @@ system: |
     {"story_size": "<preset>", "branching_depth": "<depth>", "estimated_passages": <int>, "target_word_count": <int>}
     Valid story_size: "vignette", "short", "standard", "long".
     Match branching_depth to size: vignette="light", short="light", standard="moderate", long="heavy".
-    If the brief says "vignette" → story_size="vignette", estimated_passages=10, target_word_count=3000.
-    If the brief says "short" → story_size="short", estimated_passages=20, target_word_count=10000.
-    If no size is mentioned → story_size="standard", estimated_passages=45, target_word_count=20000.
+    If the brief says "vignette" → story_size="vignette", estimated_passages=10, target_word_count=3500.
+    If the brief says "short" → story_size="short", estimated_passages=22, target_word_count=10000.
+    If the brief says "long" → story_size="long", estimated_passages=90, target_word_count=45000.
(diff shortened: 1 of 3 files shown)


pvliesdonk:
Addressed review feedback in 7936db9:

- **Claude #3**: Renamed `test_optional_scope_is_none_by_default` → `test_scope_defaults_when_omitted` with updated docstring.
- **Gemini**: Added missing `long` mapping to serialize prompt (90 passages, 45000 words).
- **Claude #1**: Aligned prompt values with PRESETS midpoints: vignette→3500 words, short→22 passages, standard→22500 words.
- **Claude #2**: The model default `target_word_count=20000` is intentionally left as-is — it's within the standard range (15000-30000) and changing it would require updating ~10 tests. The PRESETS are what enforce range validation at runtime; the model default is just a starting point for the LLM when no size is specified. The prompt now correctly guides to 22500 for "standard".

PR #455, opened 2026-01-31
fix(seed): surface CROSS_REFERENCE errors in retry feedback loop

pvliesdonk:
## Problem

SEED retry loop silently dropped `CROSS_REFERENCE` validation errors, causing two persistent failure modes that never self-corrected across 6 retry attempts:

- Default answer placed in "unexplored" instead of "explored" (check 11d)
- Paths missing a beat with `effect="commits"` (check 13)

Root cause: three compounding gaps in the error feedback pipeline.

## Changes

- **`format_semantic_errors_as_content()`** (mutations.py) — added `CROSS_REFERENCE` category handling so outer retry feedback includes "bucket misplacement" errors with their already-descriptive issue text
- **`_format_section_corrections()`** (serialize.py) — added `CROSS_REFERENCE` branch that produces "MOVE TO EXPLORED" instructions instead of falling through to the SEMANTIC handler which misleadingly says "X is INVALID, use Y"
- **Beat retry** (serialize.py) — formats beat-specific errors via `_format_section_corrections()` and appends them to the per-path prompt, so the LLM receives feedback about missing commits beats instead of re-running the identical prompt

## Not Included / Future PRs

- Prompt-level improvements to `serialize_seed.yaml` (already has strong commits instructions)
- Valid IDs context explanation for `(default)` label meaning

## Test Plan

- `uv run mypy src/questfoundry/graph/mutations.py src/questfoundry/agents/serialize.py` — passes
- `uv run ruff check` — passes
- `uv run pytest tests/unit/test_mutations.py -x -q` — 153 passed
- `uv run pytest tests/unit/test_serialize*.py -x -q` — 52 passed
- Pre-commit hooks pass

## Risk / Rollback

Low risk — changes only affect error feedback strings sent to the LLM during retry. No schema changes, no graph mutations, no new dependencies. Safe to revert by reverting this single commit.

### DIFF
--- a/src/questfoundry/agents/serialize.py
+++ b/src/questfoundry/agents/serialize.py
@@ -1060,7 +1060,7 @@ def _format_section_corrections(errors: list[SeedValidationError]) -> str:
 
         # Handle CROSS_REFERENCE errors (bucket misplacement, answer not in explored)
         if category == SeedErrorCategory.CROSS_REFERENCE:
-            cross_ref_items.append(f"- MOVE TO EXPLORED: {error.issue}")
+            cross_ref_items.append(f"- MOVE '{error.provided}' TO EXPLORED. Reason: {error.issue}")
             continue
 
         # Handle COMPLETENESS errors (missing decisions)
--- a/src/questfoundry/agents/serialize.py
+++ b/src/questfoundry/agents/serialize.py
@@ -1053,8 +1053,15 @@ def _format_section_corrections(errors: list[SeedValidationError]) -> str:
     corrections: list[str] = []
     missing_items: list[str] = []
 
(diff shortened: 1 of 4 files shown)


pvliesdonk:
Addressed review feedback in abec630:

- **Gemini**: Adopted suggestion to include `error.provided` in CROSS_REFERENCE corrections for more explicit `MOVE 'X' TO EXPLORED. Reason: ...` format.
- **Codecov**: Added 4 new tests covering CROSS_REFERENCE handling in both `format_semantic_errors_as_content()` and `_format_section_corrections()`, including a negative test verifying CROSS_REFERENCE errors don't produce misleading "is INVALID" substitutions.

Issue #130, opened 2026-01-14
Ollama truncates prompts to 4096 tokens regardless of model capability


pvliesdonk:
## Problem

When using Ollama with models that have large context windows (e.g., `gpt-oss:120b`), prompts are being truncated to 4096 tokens:

```
ai-ollama  | time=2026-01-14T17:06:04.640Z level=WARN source=runner.go:186 msg="truncating input prompt" limit=4096 prompt=4424 keep=4 new=4096
```

This happens because Ollama defaults to 4096 tokens unless `num_ctx` is explicitly set.

## Broader Need: Context Budget Awareness

Beyond fixing truncation, we need context window size programmatically available so later stages can:
- Inject more or less prior context based on available space
- Make intelligent decisions about what to include in prompts
- Avoid silent truncation that degrades output quality

Context varies wildly across providers:
| Provider | Typical Context |
|----------|----------------|
| Ollama (varies) | 8k - 128k |
| OpenAI gpt-4o | 128k |
| Anthropic Claude | 200k |
| Gemini | 1M+ |

### LangChain ChatOllama
- Supports `num_ctx` parameter directly
- Default is 2048 (LangChain) or 4096 (Ollama server)
- Docs: https://reference.langchain.com/python/integrations/langchain_ollama/

### Ollama API
- `/api/show` returns model metadata including context length
- Can query at runtime for Ollama models

### OpenAI/Anthropic
- No runtime API to query context - must use known values
- Well-documented per model

### 1. Add `ModelInfo` dataclass

```python
@dataclass
class ModelInfo:
    """Model capabilities and limits."""
    context_window: int  # Max tokens
    supports_tools: bool = True
    supports_vision: bool = False
```

### 2. Provider returns `ModelInfo` alongside model

```python
def create_chat_model(...) -> tuple[BaseChatModel, ModelInfo]:
    ...
```

Or attach to model as metadata.

### 3. Known context windows (fallback/override)

```python
KNOWN_CONTEXT_WINDOWS = {
    "openai": {
        "gpt-4o": 128_000,
        "gpt-4o-mini": 128_000,
        "gpt-4-turbo": 128_000,
    },
    "anthropic": {
        "claude-3-5-sonnet": 200_000,
        "claude-3-opus": 200_000,
    },
    "ollama": {
        "qwen3:8b": 32_768,
        # Query /api/show for unknown models
    },
}
```

### 4. Ollama: Query or configure

For Ollama, either:
[...]

PR #473, opened 2026-01-31
fix: Ollama concurrency control and smart model eviction

pvliesdonk:
## Problem

Three issues when running QuestFoundry with Ollama:

1. **Server overload**: `_serialize_beats_per_path()` fires `asyncio.create_task()` for all paths simultaneously with no concurrency limit. With 6+ paths, this floods Ollama with concurrent GPU inference requests, overwhelming the server and causing hangs/crashes.

2. **VRAM exhaustion**: When different Ollama models are used for discuss/summarize/serialize phases (hybrid provider configuration), the outgoing model stays loaded in VRAM for 5 minutes (Ollama's default `keep_alive`). Both models compete for GPU memory.

3. **Hardcoded num_ctx**: Context window is hardcoded to 32768 regardless of model capability. Models with different configured context windows get the wrong value. (#130)

### Fix 1: Concurrency limiter
- Add `asyncio.Semaphore(max_concurrency=2)` to `_serialize_beats_per_path()` in `serialize.py`
- All paths still get tasks created, but only `max_concurrency` run simultaneously
- Default of 2 allows pipelining without flooding Ollama (whose `OLLAMA_NUM_PARALLEL` defaults to 1)

### Fix 2: Smart model eviction
- Add `unload_ollama_model()` utility that sends `POST /api/generate` with `keep_alive=0` to immediately unload a model from VRAM
- Orchestrator compares resolved provider strings across discuss/summarize/serialize phases
- When models differ, builds async unload hooks; when same model, hooks are no-ops
- Stages (Dream, Brainstorm, Seed, Dress) call hooks between phase transitions
- GROW/FILL stages unaffected (single-model, no three-phase pattern)

### Fix 3: Dynamic num_ctx from Ollama
- Add `_query_ollama_num_ctx()` that queries Ollama's `/api/show` endpoint at model creation time
- Extracts `num_ctx` from the model's Modelfile `parameters` field (the configured value)
- Falls back to architecture `context_length` from `model_info` if parameters lacks num_ctx
- Falls back to 32768 if query fails (Ollama unreachable, model not found)
- Explicit `num_ctx` kwarg still takes precedence over the query

### Edge cases handled
- Same model all phases: no unload calls (no-op hooks)
- Ollama→OpenAI switch: unloads the Ollama model
- OpenAI→Ollama: no unload needed
- Non-Ollama providers: utility silently returns
- Unload failure: logged as warning, non-fatal
- Ollama unreachable for /api/show: falls back to 32768

## Not Included / Future PRs
- Server-side `OLLAMA_NUM_PARALLEL` configuration guidance (docs)
- Making `max_concurrency` configurable via project.yaml

## Test Plan
- `uv run pytest tests/unit/test_provider_factory.py -x -q` — 71 tests pass (includes 4 unload tests + 5 num_ctx query tests)
- `uv run pytest tests/unit/test_serialize.py::TestSerializeBeatsPerPathConcurrency -x -q` — 1 test passes
- `uv run mypy src/` and `uv run ruff check src/` clean

## Risk / Rollback
- Semaphore default of 2 is conservative; can be increased if needed
- Unload hooks are no-ops when all phases use the same provider (most common case)
- Unload failures are non-fatal (warning logged, pipeline continues)
- num_ctx query adds one sync HTTP call at model creation; 10s timeout with graceful fallback

### DIFF
--- a/src/questfoundry/agents/serialize.py
+++ b/src/questfoundry/agents/serialize.py
@@ -652,10 +652,14 @@ async def _serialize_beats_per_path(
     max_retries: int,
     callbacks: list[BaseCallbackHandler] | None,
     on_phase_progress: PhaseProgressFn | None = None,
+    max_concurrency: int = 2,
 ) -> tuple[list[dict[str, Any]], int]:
-    """Serialize beats for all paths in parallel.
+    """Serialize beats for all paths with bounded concurrency.
 
-    Uses asyncio.gather() to run per-path serialization concurrently.
+    Creates tasks for all paths but limits how many run simultaneously
+    using an asyncio.Semaphore.  Default of 2 prevents flooding Ollama
+    (whose OLLAMA_NUM_PARALLEL defaults to 1) while still allowing
+    pipelining.
 
     Args:
         model: Chat model to use.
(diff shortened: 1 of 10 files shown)


pvliesdonk:
## Addressing review comments

**Re: orchestrator closure scope (line 516)** — This is not a bug. Both `unload_after_discuss` and `unload_after_summarize` are local variables defined inside `run_stage()` at lines 503-504. They're initialized to `_noop` on every call, then conditionally redefined. No state leaks between `run_stage()` invocations since Python scopes these to the function body, not the instance.

**Re: seed.py unload timing (line 391)** — Acknowledged as non-blocking. Current approach is correct — unloading happens before first serialize attempt. Moving it earlier would save minimal VRAM time since the gap between summarize completion and first serialize is negligible.

**Re: factory.py broad exception (line 307)** — Agreed, keeping as-is. Cleanup operations should not break the pipeline.

All CI checks green (codecov/patch is coverage threshold, non-blocking).

Issue #479, opened 2026-02-01
Investigate switching all structured output schemas to required-only fields + JSON_MODE


pvliesdonk:
## Problem

OpenAI's `with_structured_output(method="function_calling")` (TOOL strategy) allows models to omit required fields when preceding fields are verbose. This caused DRESS Phase 1 failures where gpt-5-mini omitted `caption` from `IllustrationBrief` ~50% of the time.

The root cause is that OpenAI uses TOOL strategy because its strict JSON schema mode requires all properties in `required`, which breaks schemas with optional/defaulted fields. However, in practice the LLM always provides these "optional" fields — the defaults exist for programmatic convenience, not because the LLM should omit them.

## Proposal

Investigate whether **all** structured output schemas across stages (DREAM, BRAINSTORM, SEED, GROW, FILL, DRESS) can:

1. **Remove defaults** from fields that the LLM should always provide — making them required in the JSON schema
2. **Switch OpenAI to JSON_MODE** (`json_schema` strict mode) globally, which enforces complete schemas

This was done successfully for DRESS Phase 1 (`IllustrationBrief`, `DressPhase1Output`) as a targeted fix. The question is whether it generalizes.

## Investigation checklist

- [ ] Audit all Pydantic models used with `with_structured_output()` for fields with defaults
- [ ] For each defaulted field, determine: does the LLM always provide it? Is the default only for programmatic use?
- [ ] Identify any schemas where optional fields are genuinely optional in LLM output
- [ ] Test OpenAI JSON_MODE with the modified schemas on each stage
- [ ] Verify Ollama (already JSON_MODE) is unaffected by removing defaults

## Context

- Current strategy selection: `src/questfoundry/providers/structured_output.py`
- Original reason for TOOL on OpenAI: commit `e1fdc9a` — strict mode requires all properties in `required`
- Original reason for JSON_MODE on Ollama: PR #117 — TOOL returns None for complex nested schemas
- Targeted DRESS fix: made `IllustrationBrief` and `DressPhase1Output` all-required, passed `strategy=JSON_MODE` explicitly

pvliesdonk:
## Investigation Results

Full audit of all Pydantic models used as schemas for `with_structured_output()` across all stages (DREAM, BRAINSTORM, SEED, GROW, FILL, DRESS).

---

### Current Strategy Selection (`providers/structured_output.py`)

| Provider | Default Strategy | Reason |
|----------|-----------------|--------|
[...]

PR #494, opened 2026-02-01
fix(dress): change style_overrides from dict to str for OpenAI compat

pvliesdonk:
## Problem
DRESS Phase 1 fails on OpenAI (gpt-5) with:
```
Invalid schema for response_format 'DressPhase1Output':
'required' is required to be supplied and to be an array including every key in properties.
Extra required key 'style_overrides' supplied.
```

Root cause: `IllustrationBrief.style_overrides` is typed `dict[str, str]`, which generates `"additionalProperties": {"type": "string"}` in the JSON schema. OpenAI strict mode only supports `additionalProperties: false` — it strips the unsupported property definition, then rejects the schema because `style_overrides` is in `required` but no longer in `properties`.

PR #480 switched DRESS briefs to JSON_MODE and made all fields required, but didn't catch that `dict[str, str]` is incompatible with OpenAI's strict schema. This worked on Ollama (which accepts the schema) but fails on OpenAI.

## Changes
- Change `style_overrides` type from `dict[str, str]` to `str` on `IllustrationBrief`
- Update prompt template description to match
- Update all test fixtures from `style_overrides={}` to `style_overrides=""`

## Not Included / Future PRs
- Global audit of all schemas for OpenAI strict mode compat (#479)

## Test Plan
```bash
uv run ruff check src/questfoundry/models/dress.py src/questfoundry/pipeline/stages/dress.py  # Clean
uv run mypy src/questfoundry/models/dress.py src/questfoundry/pipeline/stages/dress.py  # Clean
uv run pytest tests/unit/test_dress_models.py tests/unit/test_dress_stage.py -x -q  # 98 passed
```

Verified schema no longer contains `additionalProperties` with type schema — all properties are simple types compatible with OpenAI strict mode.

## Risk / Rollback
- Low risk: `style_overrides` was never consumed as a dict anywhere in the codebase
- The only programmatic usage (cover image brief) already passes it as a string
- Ollama is unaffected (accepts both dict and string schemas)

Discovered investigating `projects/test-20260201T1237/` DRESS failure.


pvliesdonk:
Addressing review comments:

1. **style_overrides unused in assemble_image_prompt** — Valid pre-existing bug, out of scope for this fix. Created follow-up issue.
2. **dict[str, str] in grow.py** — Already covered by #479 audit scope.
3. **CLI test fixes** — Added in second commit to fix pre-existing CI failures from SHIP merge.

Issue #512, opened 2026-02-02
fix(dress): cover brief should target vision::main, not synthetic passage::cover


pvliesdonk:
## Problem

Cover illustrations are NOT passages — they're presentation metadata. The current implementation creates a synthetic `passage::cover` node to satisfy `apply_dress_brief()`'s requirement that briefs target passages. This causes SHIP to reject it ("passage missing prose") and pollutes the graph with synthetic nodes.

This is part of a broader architectural smell introduced in PR #506, which also created a phantom `vision::main` node for `story_title` storage (see #515 for that fix).

## Root Cause Analysis

1. `apply_dress_brief()` hardcodes `graph.ref("passage", raw_passage_id)` — forces all briefs to target passages
2. `_create_cover_brief()` creates `passage::cover` as a workaround
3. SHIP validates all passages have prose → rejects `passage::cover`
4. Every IF system (Twine, Inform, ink, ChoiceScript) treats cover art as bibliographic metadata outside the story graph, never as a narrative unit

## Design Decision: Standalone Cover Illustration

A cover illustration should be **standalone** — no `Depicts` edge, no `targets` edge. This matches IF conventions and keeps the edge contracts clean.

- `Depicts` means "art shown with a passage" — cover isn't shown with any passage
- The *absence* of a `Depicts` edge is meaningful: "this isn't a passage illustration"
- Add `"cover"` to `IllustrationCategory` — self-documenting
- `category="cover"` tells exporters to render it on the title page, not inline

### Ontology Changes (docs/design/00-spec.md)

- **IllustrationBrief `category`** (line 582): Add `cover` to category list
- **IllustrationBrief description** (line 576): Note that cover briefs have no `targets` edge
- `Depicts` and `targets` edges stay **unchanged** — passage-only

### Code Changes

| File | Change |
|------|--------|
| `pipeline/stages/dress.py` | Remove synthetic `passage::cover` creation; cover brief has no `targets` edge |
| `graph/dress_mutations.py` | Handle briefs with no `targets` edge in `apply_dress_illustration` |
| `models/dress.py` | Add `"cover"` to `IllustrationCategory` |
| `export/base.py` | Add `cover: ExportIllustration | None` to `ExportContext` |
| `export/context.py` | Extract cover illustration separately (no `Depicts` edge, `category="cover"`) |
| `export/html_exporter.py` | Render cover on title page |
| `export/twee_exporter.py` | Render cover in `StoryInit` |

### How Exports Handle It

[...]

PR #513, opened 2026-02-02
feat(dress): add progress feedback during image generation

pvliesdonk:
## Problem
`generate-images` provides no intermediate progress — users only see the final result after all distillation and rendering completes.

## Changes
- Thread `on_phase_progress` through `_phase_4_generate()` via explicit parameter and instance fallback
- Emit `"Distilling {i}/{total} prompts"` progress during Pass 1 (prompt distillation)
- Emit `"Rendering {i}/{total}"` progress during Pass 2 (image generation)
- Progress fires from both `run_generate_only()` (explicit param) and `execute()` (instance callback)

## Not Included / Future PRs
- Cover brief reanchoring (#512) — separate issue

## Test Plan
- `uv run pytest tests/unit/test_dress_stage.py::TestRunGenerateOnly -x -q` — 5 tests pass
- New test `test_progress_reports_distill_and_render` verifies both distill and render progress callbacks fire

## Risk / Rollback
- Additive change only — existing behavior unchanged when no callback is provided
- `getattr(self, "_on_phase_progress", None)` fallback is safe for all callers

### DIFF
--- a/src/questfoundry/pipeline/stages/dress.py
+++ b/src/questfoundry/pipeline/stages/dress.py
@@ -250,6 +250,7 @@ class DressStage:
         self._on_assistant_message = on_assistant_message
         self._on_llm_start = on_llm_start
         self._on_llm_end = on_llm_end
+        self._on_phase_progress = on_phase_progress
         self._summarize_model = summarize_model
         self._user_prompt = user_prompt
         self._unload_after_discuss = kwargs.get("unload_after_discuss")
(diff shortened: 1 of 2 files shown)


pvliesdonk:
**Regarding Gemini's callback refactoring suggestion:** The current design is intentional. `run_generate_only()` is a public API that explicitly passes the callback, while `execute()` stores it as an instance attribute for the phase loop. Removing the parameter would couple the public API to instance state. Keeping both paths is a deliberate choice for testability.

PR #514, opened 2026-02-02
fix(dress): entity name dereference and distiller CLIP context

pvliesdonk:
## Problem
1. **Entity names not dereferenced**: `build_image_brief()` passes anonymous visual descriptions (e.g., "tall woman in dark coat") without entity names. The distiller LLM sees character names in the `subject` field (e.g., "Eleanor") but has no way to map them to descriptions. SD has no concept of character names.

2. **Missing CLIP context**: The system prompt doesn't explain *why* distillation happens — no mention of CLIP token limits or that excess tokens are silently dropped.

## Changes
- `build_image_brief()`: entity fragments now include entity ID prefix — `"elara: tall woman, dark coat"` instead of just `"tall woman, dark coat"`
- A1111 system prompt: added CONTEXT paragraph explaining CLIP token window and silent truncation
- A1111 system prompt: added ENTITY EXPANSION rule instructing the LLM to replace character names with visual descriptions from the Entities field

## Not Included / Future PRs
- Cover brief reanchoring (#512)
- Progress feedback (#513)

## Test Plan
- `uv run pytest tests/unit/test_image_a1111.py tests/unit/test_image_brief.py -x -q` — 58 tests pass
- Updated assertions verify entity fragments include name prefix and system prompt includes CLIP/entity content

## Risk / Rollback
- Entity fragment format change is additive — existing code that consumes fragments will see richer context
- System prompt changes may slightly alter distilled output quality (intended improvement)

### DIFF
--- a/src/questfoundry/pipeline/stages/dress.py
+++ b/src/questfoundry/pipeline/stages/dress.py
@@ -1364,7 +1364,7 @@ def build_image_brief(graph: Graph, brief: dict[str, Any]) -> ImageBrief:
         if ev:
             frag = ev.get("reference_prompt_fragment", "")
             if frag:
-                entity_fragments.append(frag)
+                entity_fragments.append(f"{raw_eid}: {frag}")
 
     return ImageBrief(
         subject=brief.get("subject", ""),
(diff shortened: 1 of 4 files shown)


pvliesdonk:
Addressed Gemini's review: fixed entity ID example from "elara" to "eleanor" to match the name "Eleanor" in the prompt (bab85eb).

Issue #550, opened 2026-02-03
Add FILL prose guidance for endings and character introductions


pvliesdonk:
## Problem

The pipeline marks passages as `is_ending` and has scene-type tags (`introduce`, `develop`, etc.) but none of this metadata reaches the FILL prose template. LLMs receive no guidance on how to write endings or introduce characters.

### Impact (from 5-model benchmark, same prompt)

| Model | Introductions | Endings |
|-------|---------------|---------|
| qwen3:4b | Poor — characters appear mid-scene without description | Nonexistent — dissolves into philosophical ambiguity |
| gpt-5-mini | Excellent | Strong — clear consequence and closure |
| gpt-5 | Excellent | Strong — devastating final images |
| gemini-2.5-pro | Good | Weak — reads as chapter break, not ending |

GPT models handle this intuitively. Smaller models and Gemini need explicit guidance. The fix is prompt-only — no schema or pipeline changes required.

### 1. Thread `is_ending` into FILL prose context

The flag exists on `ExportPassage` and in the graph. Pass it to the FILL prose template so the LLM knows when it's writing a concluding passage.

**In `fill_phase1_prose.yaml`, add conditional guidance:**

```yaml
{% if is_ending %}
## Ending Passage Guidance
This is a story ending. The reader's journey concludes here.
- Resolve at least one open dramatic question from this path
- End with a concrete image or action, not abstract reflection
- The reader should feel the story has concluded, not paused
- Match the ending tone to the path's emotional arc (path_mood)
{% endif %}
```

### 2. Add character introduction guidance

The entity list is already passed to FILL. Add general guidance for first appearances without requiring first-appearance tracking:

```yaml
## Character Introductions
When a character appears in a passage for the first time:
- Ground them with one physical detail and one defining action
- Establish their relationship to the POV character through behavior, not exposition
- Give them a sensory marker (a sound, habit, or object) the reader can anchor to
```

This leverages the LLM's ability to infer whether a character has appeared before from the conversation context and passage ordering.

## Files Involved

- `prompts/templates/fill_phase1_prose.yaml` — Add ending and introduction guidance (~15-20 lines)
- `src/questfoundry/pipeline/stages/fill.py` — Thread `is_ending` flag into prose context dict (if not already present)

## Scope

~20-30 lines changed. Prompt-only fix, no schema or model changes. Zero additional LLM calls or token cost.

## Testing

1. Re-run with qwen3:4b and verify endings have closure (not philosophical dissolution)
[...]

PR #555, opened 2026-02-03
feat(fill): thread is_ending into prose generation context

pvliesdonk:
## Problem
The pipeline computes `is_ending` only at SHIP time (export layer), but FILL prose generation has no awareness of whether a passage is a story ending. This causes:
- qwen3:4b endings dissolve into philosophical ambiguity
- Gemini endings read as chapter breaks, not conclusions
- GPT-5 handles it intuitively but would still benefit from explicit guidance

Refs: #550

## Changes
- Add `compute_is_ending(graph, passage_id)` to `fill_context.py` — checks for outgoing `choice_from` edges
- Add `format_ending_guidance(is_ending)` — returns craft guidance with BAD/GOOD examples for endings, empty string for non-endings
- Thread `ending_guidance` into the FILL Phase 1 context dict in `fill.py`
- Add `{ending_guidance}` placeholder to `fill_phase1_prose.yaml` between Rules and Poly-State sections

## Not Included / Future PRs
- Character introduction guidance — needs first-appearance tracking design (separate issue)
- Gemini retry improvements — PR3 in stack
- Two-step FILL architecture — PR4 in stack

## Test Plan
- `uv run pytest tests/unit/test_fill_context.py::TestIsEnding tests/unit/test_fill_context.py::TestEndingGuidance -x -q` — 5 tests pass
- `uv run mypy src/questfoundry/graph/fill_context.py src/questfoundry/pipeline/stages/fill.py` — clean
- Pre-commit passes on all changed files

## Risk / Rollback
- `compute_is_ending` uses the same edge-checking pattern as `grow_validation.py:231`
- Empty `ending_guidance` for non-endings means zero impact on existing prose generation
- No schema or model changes

### DIFF
--- a/prompts/templates/fill_phase1_prose.yaml
+++ b/prompts/templates/fill_phase1_prose.yaml
@@ -101,6 +101,8 @@ system: |
       GOOD: "She picked up his coffee mug, still warm, and held it with both hands."
       GOOD: "His jaw tightened. He set down the glass without drinking."
 
+  {ending_guidance}
+
   ## Poly-State Examples (CRITICAL for shared beats)
   GOOD: "The stranger's expression was unreadable" (works for trust or betrayal)
   GOOD: "Something had shifted between them" (ambiguous emotional change)
--- a/src/questfoundry/graph/fill_context.py
+++ b/src/questfoundry/graph/fill_context.py
@@ -988,8 +988,12 @@ def format_path_arc_context(graph: Graph, passage_id: str, arc_id: str) -> str:
 def compute_is_ending(graph: Graph, passage_id: str) -> bool:
     """Determine if a passage is a story ending (no outgoing choices).
 
-    A passage is an ending if no choice node has a ``choice_from`` edge
-    pointing to it, meaning the reader has no further navigation options.
+    A passage is an ending if no ``choice_from`` edge has this passage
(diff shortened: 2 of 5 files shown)


pvliesdonk:
Addressed review findings in 9a24e47:

1. **Docstring clarified** — now documents the `choice_from` edge direction convention (FROM choice TO originating passage) with a reference to `grow_validation.py:212`.

2. **Removed nonexistent passage test** — Claude correctly identified this masks bugs. A nonexistent passage silently returns `True` (ending) rather than surfacing the error.

3. **Re: Gemini's "critical" edge direction claim** — this is incorrect. The `choice_from` edge convention is FROM choice TO passage (see `grow.py:1950`: `graph.add_edge("choice_from", choice_id, p_id)` and `grow_validation.py:233`: `passages_with_outgoing = {edge["to"] for edge in choice_from_edges}`). So `get_edges(to_id=passage_id)` correctly finds choices originating FROM this passage. The implementation and test are correct.

4. **Template spacing** — the `{ending_guidance}` placeholder produces either a full guidance block or an empty string. Empty string between sections is standard for this template (same pattern as `{vocabulary_note}` above). No change needed.

All 87 tests pass, pre-commit green.

Issue #549, opened 2026-02-03
Gemini 2.5 Pro underperforms expectations in FILL prose quality


pvliesdonk:
## Problem

Gemini 2.5 Pro scores 8.5/10 on prose quality — only 0.5 above qwen3:4b (a 4B local model) and below both GPT-5 (9.2) and GPT-5-mini (9.0). For a frontier model, this is unexpectedly low.

### Benchmark Results (test-20260202 runs, same prompt)

| Model | Prose Quality | Total Tokens | Wall Time | FILL Retries |
|-------|-------------|-------------|-----------|-------------|
| gpt-5 | 9.2/10 | 595K | 105min | 2 |
| gpt-5-mini | 9.0/10 | 503K | 45min | 2 |
| **gemini-2.5-pro** | **8.5/10** | **469K** | **28min** | **11** |
| qwen3:4b | 8.0/10 | 340K | 10-17min | ~3 |

### Observed Quality Issues (T2208 "Where the Rain Washes Nothing Clean")

- Occasional thematic over-statement (telling theme instead of showing)
- Limited female character voice differentiation
- Sensory palette doesn't evolve in later passages (rain/smoke/whiskey repeated without novelty)
- Some adverb tag creep in dialogue ("voice a low rasp")
- 3 instances of "stale air" as default beat-closing device

These are the kinds of issues you'd expect from a model fighting its way through repair loops rather than flowing creatively.

### 1. Excessive FILL retries (11 vs 2 for GPT-5)

Gemini needed 11 revision passes in FILL. Each retry shifts the model from creative prose into error-correction mode. By the time output validates, the prose has been "corrected" into something more mechanical. The repair loop is a quality tax.

**Action:** Analyze `projects/test-20260202T2208/logs/llm_calls.jsonl` to determine:
- What validation errors triggered retries (schema-structural vs content-quality?)
- Whether prose quality degrades across retry iterations
- Whether the serialization prompt needs Gemini-specific adjustments

### 2. Prompts tuned for OpenAI/Ollama, not Gemini

The entire prompt corpus was developed and iterated against OpenAI and qwen3. Gemini may interpret "sparse register," "punchy rhythm," and avoid-list instructions differently. Small interpretation differences compound across 15+ passages.

**Action:** Compare the raw LLM call content between T2208 (Gemini) and T2125 (GPT-5) for the same FILL phase — are the prompts identical? Does Gemini need different phrasing?

[...]

PR #556, opened 2026-02-03
fix(fill): improve retry diagnostics, prose-preservation feedback, and null-stripping

pvliesdonk:
## Problem
Gemini 2.5 Pro needed 11 FILL retries vs 2 for GPT-5 (#549). Two root causes addressed:
1. Retry feedback told the model to "fix errors and try again" without distinguishing structural issues (JSON shape) from content issues (field constraints). This caused the model to rewrite good prose into something safer.
2. Gemini often emits explicit `null` for optional fields, which Pydantic rejects when the field type is `str` (not `str | None`).

## Changes
- Add `_classify_validation_error()` to categorize failures as "structural" (missing fields, wrong types) vs "content" (min_length, enum mismatch)
- Add structured logging with `failure_type`, `fields_missing`, `fields_invalid` to the retry warning
- Rewrite `_build_error_feedback` to include prose-preservation language for structural errors: "Keep your prose intact. Fix ONLY the structural issue."
- Apply `strip_null_values()` in `unwrap_structured_result` for parsed dict results, reusing existing function from `artifacts/validator.py`

## Not Included / Future PRs
- Two-step FILL architecture — PR4 in stack
- Gemini-specific temperature curve — deferred pending retry count data with these fixes

## Test Plan
- `uv run pytest tests/unit/test_fill_stage.py::TestClassifyValidationError tests/unit/test_fill_stage.py::TestBuildErrorFeedback -x -q` — 5 tests pass
- `uv run pytest tests/unit/test_structured_output.py::TestUnwrapStructuredResult -x -q` — 6 tests pass
- `uv run mypy src/questfoundry/pipeline/stages/fill.py src/questfoundry/providers/structured_output.py` — clean

## Risk / Rollback
- Error classification is additive — existing retry behavior preserved for non-structural errors
- Null-stripping is provider-agnostic and matches the existing `strip_null_values` pattern already used in `serialize.py:316`
- Prose-preservation language only activates for structural errors; content errors get the existing generic message

### DIFF
--- a/src/questfoundry/pipeline/stages/fill.py
+++ b/src/questfoundry/pipeline/stages/fill.py
@@ -92,7 +92,9 @@ _SLIDING_WINDOW_SIZES: dict[str, int] = {
 }
 
 
-_STRUCTURAL_ERROR_TYPES = frozenset({"missing", "value_error", "type_error", "extra_forbidden"})
+_STRUCTURAL_ERROR_TYPES = frozenset(
+    {"missing", "missing_argument", "type_error", "extra_forbidden"}
+)
 
 
 def _classify_validation_error(
@@ -124,12 +126,12 @@ def _classify_validation_error(
         else:
             invalid.append(field_path)
 
-    if missing and not invalid:
+    # Structural errors take precedence: fix missing/wrong-type fields first
+    # since content validation cannot run on absent fields.
+    if missing:
         return ("structural", missing, invalid)
-    if invalid and not missing:
+    if invalid:
         return ("content", missing, invalid)
-    if missing and invalid:
-        return ("structural", missing, invalid)
     return ("unknown", missing, invalid)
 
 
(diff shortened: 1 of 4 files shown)


pvliesdonk:
Addressed all review findings in 79e12db:

1. **Removed `value_error` from `_STRUCTURAL_ERROR_TYPES`** — too broad for Pydantic v2. Added `missing_argument` instead (Claude's finding).
2. **Simplified classification logic** — structural takes precedence when both error types exist, with a comment explaining why (fix missing fields first, content validation can't run on absent fields). Reduced from 4 branches to 3 (Gemini's suggestion + Claude's documentation request).
3. **Made `_build_error_feedback` a `@staticmethod`** — doesn't use `self` (Gemini's suggestion). Cleaned up tests accordingly.
4. **Added mixed error test** — validates behavior when both structural and content errors exist (Claude's missing test case).

All 68 tests pass, pre-commit green.

PR #568, opened 2026-02-03
feat(seed): add beat uniqueness validation and low-count advisory

### LINKED ISSUE #178: feat: add beat count validation to SEED schema
## Problem

The SEED stage should produce a specific number of beats per thread (typically 4 beats × 4 threads = 16 beats total). Currently there's no schema-level validation to ensure the LLM produces the expected beat count.

From #172 analysis, GPT-4o under-produces beats (6 instead of 16) due to RLHF brevity optimization. Schema validation would catch this early and trigger repair loops.

## Proposed Solution

Add validation constraints to the SEED artifact schema:

```json
{
  "threads": {
    "type": "array",
    "minItems": 4,
    "maxItems": 4,
    "items": {
      "properties": {
        "beats": {
          "type": "array",
          "minItems": 4,
          "maxItems": 4
        }
      }
    }
  }
}
```

Or implement as a custom validator in the SEED stage that checks beat counts and provides specific feedback to the LLM for repair.

## Acceptance Criteria

- [ ] SEED validation fails if beat count doesn't match expectations
- [ ] Error message clearly indicates expected vs actual beat count
- [ ] Repair loop can use this feedback to regenerate with correct count
- [ ] Configurable expected counts (not hardcoded)

## References

- Extracted from #172 (Model Selection Strategy)
- Key finding: "GPT-4o produced 6 beats vs 16 expected"

pvliesdonk:
## Problem
The SEED stage lacks validation for duplicate beat IDs within a path's beats, which could cause graph conflicts in GROW. Additionally, some models (notably GPT-4o) under-produce beats due to brevity optimization, producing 6 beats instead of the expected 16 (4 per path × 4 paths). There's no advisory feedback when this happens.

## Changes
- **`models/seed.py`**: Add `model_validator` to `PathBeatsSection` that rejects duplicate beat IDs with a descriptive error message
- **`pipeline/stages/seed.py`**: Add advisory warning when average beats/path < 3 (expected ~4), logged as `seed_low_beat_count` event
- **`tests/unit/test_seed_stage.py`**: Add `TestPathBeatsSectionValidation` class with 5 tests:
  - `test_valid_four_beats` — 4 unique beats passes
  - `test_valid_two_beats` — 2 beats (minimum) passes
  - `test_one_beat_rejected` — fewer than 2 rejected by min_length
  - `test_five_beats_rejected` — more than 4 rejected by max_length
  - `test_duplicate_beat_ids_rejected` — duplicate IDs rejected by model_validator

## Not Included / Future PRs
- Configurable expected beats per path (currently hardcoded thresholds)
- Beat count as a repair loop trigger (currently advisory only)

## Test Plan
```bash
uv run ruff check src/ && uv run mypy src/questfoundry/models/seed.py src/questfoundry/pipeline/stages/seed.py  # Pass
uv run pytest tests/unit/test_seed_stage.py -x -q  # 26 passed (5 new)
```

## Risk / Rollback
Low. The duplicate ID validator is a new constraint but existing SEED output from healthy models always produces unique beat IDs. The advisory warning is log-only, no behavior change.

Closes #178

### DIFF
--- a/src/questfoundry/models/seed.py
+++ b/src/questfoundry/models/seed.py
@@ -342,6 +342,19 @@ class PathBeatsSection(BaseModel):
         description="2-4 initial beats for this specific path",
     )
 
+    @model_validator(mode="after")
+    def _check_beat_ids_unique(self) -> PathBeatsSection:
+        """Validate that beat IDs are unique within this path's beats."""
+        ids = [b.beat_id for b in self.initial_beats]
+        dupes = [bid for bid in ids if ids.count(bid) > 1]
+        if dupes:
+            msg = (
+                f"Duplicate beat IDs found: {sorted(set(dupes))}. "
+                f"Each beat must have a unique beat_id."
+            )
+            raise ValueError(msg)
+        return self
+
 
 class ConvergenceSection(BaseModel):
     """Wrapper for serializing convergence sketch separately."""
--- a/src/questfoundry/models/seed.py
+++ b/src/questfoundry/models/seed.py
@@ -345,12 +345,13 @@ class PathBeatsSection(BaseModel):
     @model_validator(mode="after")
     def _check_beat_ids_unique(self) -> PathBeatsSection:
         """Validate that beat IDs are unique within this path's beats."""
+        from collections import Counter
+
         ids = [b.beat_id for b in self.initial_beats]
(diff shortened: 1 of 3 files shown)


pvliesdonk:
Addressed review feedback in 3e3ae01: switched to `collections.Counter` for O(n) duplicate detection as suggested.

PR #571, opened 2026-02-03
feat(config): support fill.two_step in project.yaml

### LINKED ISSUE #565: feat(config): support fill.two_step in project.yaml
Allow configuring two-step prose generation default in project.yaml:

```yaml
fill:
  two_step: true  # default
```

Currently two-step is controlled only via CLI flag (`--two-step/--no-two-step`). Adding project config support would let users set a per-project default without repeating the flag.

Deferred from PR #560.

pvliesdonk:
## Problem
The `--two-step` CLI flag for FILL stage prose generation has no project-level configuration, requiring users to pass it every time they run `qf fill` or `qf run`.

## Changes
- Add `FillConfig` dataclass with `two_step: bool = True` field and `from_dict()` class method
- Add `fill: FillConfig` field to `ProjectConfig`, parsed from `fill:` section in `project.yaml`
- Change CLI `--two-step` parameters from `bool` (default `True`) to `bool | None` (default `None`) so the orchestrator can distinguish explicit CLI override from config default
- Orchestrator resolves: CLI flag (if set) > `config.fill.two_step` > hardcoded default (`True`)
- Add 7 tests for `FillConfig` parsing and `ProjectConfig` integration

Example `project.yaml`:
```yaml
name: my-story
fill:
  two_step: false
```

## Not Included / Future PRs
- No additional fill config fields (max_concurrency, etc.) — can be added later as needed

## Test Plan
```
uv run ruff check src/questfoundry/pipeline/config.py src/questfoundry/cli.py  # All checks passed
uv run mypy src/questfoundry/pipeline/config.py src/questfoundry/pipeline/orchestrator.py src/questfoundry/cli.py  # no issues
uv run pytest tests/unit/test_config.py -x -q  # 48 passed
uv run pytest tests/unit/test_cli.py -x -q  # 78 passed
```

## Risk / Rollback
- Low risk — when `fill:` section is absent from project.yaml, behaviour is identical to before
- CLI flag still works and takes precedence over config

Closes #565

### DIFF
--- a/src/questfoundry/cli.py
+++ b/src/questfoundry/cli.py
@@ -461,7 +461,7 @@ def _run_stage_command(
     resume_from: str | None = None,
     image_provider: str | None = None,
     image_budget: int = 0,
-    two_step: bool = True,
+    two_step: bool | None = None,
     language: str | None = None,
 ) -> None:
     """Common logic for running a stage command.
@@ -516,7 +516,7 @@ def _run_stage_command(
         context["image_provider"] = image_provider
     if image_budget > 0:
         context["image_budget"] = image_budget
-    if stage_name == "fill":
+    if stage_name == "fill" and two_step is not None:
         context["two_step"] = two_step
     if language:
         context["language"] = language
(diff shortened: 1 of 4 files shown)


pvliesdonk:
Addressed review feedback in 91c90fd: wrapped `two_step` resolution in `if stage_name == "fill":` to avoid resolving fill-specific config for other stages.

PR #614, opened 2026-02-04
fix(inspect): emit valid JSON and correct start detection

pvliesdonk:
## Problem
`qf inspect --json` is not machine-readable because Rich wraps long lines at terminal width, splitting string values and producing invalid JSON. Also, start passage counting in both `qf inspect` and exports incorrectly treats hub-spoke return links (`is_return=True`) as incoming, leading to `start_passages: 0` and potentially wrong start passage selection.

## Changes
- Emit JSON with Rich `soft_wrap=True` to prevent line-wrapping inside string values.
- Exclude `is_return` choices from incoming-edge accounting when marking start passages in export context.
- Exclude `is_return` choices from incoming-edge accounting for `qf inspect` branching stats.
- Add unit tests covering valid JSON output and return-link start detection.

## Not Included / Future PRs
- No changes to GROW choice `requires/grants` semantics or linear-stretch collapsing (tracked separately).

## Test Plan
- `uv run ruff check src/questfoundry/cli.py src/questfoundry/export/context.py src/questfoundry/inspection.py tests/unit/test_cli.py tests/unit/test_export_context.py tests/unit/test_inspection.py`
- `uv run pytest tests/unit/test_cli.py -x -q`
- `uv run pytest tests/unit/test_export_context.py -x -q`
- `uv run pytest tests/unit/test_inspection.py -x -q`

## Risk / Rollback
- Low risk: changes are isolated to CLI output and start/ending detection logic.
- Rollback: revert this PR to restore previous behavior (JSON output + start counting).

### DIFF
--- a/src/questfoundry/cli.py
+++ b/src/questfoundry/cli.py
@@ -2556,7 +2556,14 @@ def inspect(
         import dataclasses
         import json
 
-        console.print(json.dumps(dataclasses.asdict(report), indent=2))
+        # Emit machine-readable JSON. Rich will otherwise wrap long lines at the
+        # terminal width (80), which can split string values and break JSON.
+        console.print(
+            json.dumps(dataclasses.asdict(report), indent=2),
+            soft_wrap=True,
+            markup=False,
+            highlight=False,
+        )
         return
 
     _render_inspection_report(report)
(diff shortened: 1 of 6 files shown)


pvliesdonk:
Addressed the review suggestion in `src/questfoundry/inspection.py`: switched `has_incoming` to a set comprehension and use `choice_from` edges for `has_outgoing`. Pushed commit `293cdac`.

PR #622, opened 2026-02-04
fix(grow): write topology artifact to grow.yaml

pvliesdonk:
## Problem
`artifacts/grow.yaml` currently contains telemetry (counts + phase results) instead of the GROW topology artifact (arcs/beats/passages/choices/codewords). This is inconsistent with other stages and with `docs/design/00-spec.md` (GROW output should be the topology with passage summaries, not prose).

## Changes
- Make `GrowStage.execute()` return the topology artifact produced by `extract_grow_artifact()` (orchestrator writes it).
- Improve GROW artifact extraction to preserve arc `sequence` order and include key beat/passage/choice fields.
- Add unit tests for GROW artifact extraction and update GROW stage unit tests.

## Not Included / Future PRs
- No new semantic continuity checks added to `qf inspect` (analysis only in this discussion).

## Test Plan
- `uv run pytest tests/unit/test_grow_stage.py tests/unit/test_grow_enrichment.py -x -q`

## Risk / Rollback
- Low risk: affects artifact contents and stage return value; graph mutation behavior is unchanged.
- Rollback: revert this PR to restore previous grow.yaml telemetry output.

### DIFF
--- a/src/questfoundry/artifacts/enrichment.py
+++ b/src/questfoundry/artifacts/enrichment.py
@@ -253,7 +253,7 @@ def _extract_arcs(graph: Graph) -> list[dict[str, Any]]:
         entry: dict[str, Any] = {
             "arc_id": arc_id,
             "arc_type": data.get("arc_type", "branch"),
-            "paths": data.get("paths", []),
+            "paths": sorted(data.get("paths", []) or []),
             "sequence": beat_sequence,
         }
         arcs.append(entry)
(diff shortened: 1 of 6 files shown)


pvliesdonk:
Addressed Gemini review notes and fixed CI failures: updated integration/unit tests to assert the new GROW topology artifact shape (arcs/beats/passages/choices/codewords) instead of telemetry, and made artifact output more deterministic by sorting ID lists where order is irrelevant.

pvliesdonk:
Added more unit coverage for GROW artifact extraction (sorting + fallbacks: arc paths, arc_contains fallback, beat/passsage entity sorting, codeword edge fallback) to address codecov/patch failure.

PR #623, opened 2026-02-04
feat(fill): warn on hard transitions to smooth cuts

pvliesdonk:
## Problem
FILL has the prose context needed to smooth abrupt, low-overlap transitions (“hard cuts”), but today the prompt doesn’t explicitly call them out. This can leave continuity work to chance, especially when arcs weave multiple paths.

## Changes
- Add `format_continuity_warning()` to detect likely hard cuts from the previous passage in the current arc.
- Inject `continuity_warning` into the FILL prose-generation context.
- Update `fill_phase1_prose.yaml` and `fill_phase1_prose_only.yaml` to include a dedicated Continuity Warning section.
- Add unit tests for the heuristic.

## Not Included / Future PRs
- No automatic graph mutations to fix hard cuts (this is a prompt-level guardrail only).

## Test Plan
- `uv run pytest tests/unit/test_fill_continuity_warning.py -x -q`

## Risk / Rollback
- Low risk: only affects prompt context. If it causes unwanted behavior, revert this PR.


pvliesdonk:
Addressed Gemini review feedback: expanded unit test coverage to cover additional suppression conditions (shared location, micro_beat, intersection hint, out-of-bounds index). This should also improve patch coverage for the new heuristic.

PR #628, opened 2026-02-05
fix(grow): start prune BFS from synthetic prologue when it exists

### LINKED ISSUE #627: bug(grow): prune reachability must start from synthetic prologue (real start)
## Problem
When GROW Phase 9 creates a synthetic `passage::prologue` (because there are multiple orphan starts), Phase 11 prune still computes reachability starting from the first *spine* passage. This can incorrectly prune the prologue itself and any other paths that are only reachable via the prologue (e.g. sibling branch arcs), leaving a graph with missing/invalid starting structure.

We saw this in `projects/test-20260204T2106`: SEED produced 2 major paths (`touch` + `talk`). GROW created a synthetic prologue (the phase output even said "with synthetic prologue"), but after prune only the touch path remained. The graph also contained a choice whose `from_passage` was `passage::prologue` while `passage::prologue` had been deleted.

## Expected
If a synthetic prologue is created, it is the **real** story start and must be treated as the BFS root for reachability. Prune should keep the prologue and all passages reachable from it (including both sibling start paths).

## Actual
Prune uses the first spine beat’s passage as the BFS root, so:
- `passage::prologue` can be considered unreachable and removed.
- Any non-spine start paths can be pruned even though they’re intended to be reachable via the prologue.
- The graph can end up with dangling choice references (choices pointing from `passage::prologue` that no longer exists).

## Where
`src/questfoundry/pipeline/stages/grow.py`
- `_phase_9_choices` creates `passage::prologue` when `len(orphan_starts) > 1`.
- `_phase_11_prune` → `_reachable_via_choices` currently finds the start passage by scanning for the spine arc’s first beat.

## Suggested fix
In `_reachable_via_choices`, if `passage::prologue` exists (or more generally: if there are multiple orphan starts / a synthetic prologue was created), start BFS from the prologue passage instead of the first spine passage.

Alternative: compute reachability starting from all orphan starts, but this should still preserve the “single entrypoint” intent of the prologue.

## Repro notes
- Run SEED with 1 dilemma fully explored (2 paths) so GROW creates multiple orphan starts.
- Run GROW; observe that after prune, only one path may remain and/or prologue is missing.

pvliesdonk:
## Problem
When GROW Phase 9 creates a synthetic `passage::prologue` (because there are multiple orphan starts), Phase 11 prune still computes reachability starting from the first spine passage. This incorrectly prunes the prologue itself and any paths only reachable via the prologue.

Closes #627

## Changes
- Modified `_reachable_via_choices` in `grow.py` to check for `passage::prologue` first
- If prologue exists, use it as BFS start (it's the real story start)
- Otherwise, fall back to existing spine passage logic
- Added test `test_prune_starts_from_prologue_when_exists` covering the scenario

## Not Included / Future PRs
- None

## Test Plan
```bash
uv run pytest tests/unit/test_grow_algorithms.py::TestPhase11Integration -xvs
# 6 passed
```

All existing Phase 11 tests still pass, plus the new test verifies the fix.

## Risk / Rollback
- Low risk: only changes BFS start selection logic
- Backward compatible: existing behavior unchanged when no prologue exists
- Rollback: revert this PR

### DIFF
--- a/src/questfoundry/pipeline/stages/grow.py
+++ b/src/questfoundry/pipeline/stages/grow.py
@@ -117,6 +117,7 @@ class GrowStage:
         self._lang_instruction: str = ""
 
     CHECKPOINT_DIR = "snapshots"
+    PROLOGUE_ID = "passage::prologue"
 
     # Type for async phase functions: (Graph, BaseChatModel) -> GrowPhaseResult
     PhaseFunc = Callable[["Graph", "BaseChatModel"], Awaitable[GrowPhaseResult]]
@@ -2231,7 +2232,7 @@ class GrowStage:
                 orphan_count=len(orphan_starts),
                 orphans=orphan_starts[:5],
             )
-            prologue_id = "passage::prologue"
+            prologue_id = self.PROLOGUE_ID
             graph.create_node(
                 prologue_id,
                 {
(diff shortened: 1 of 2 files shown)


pvliesdonk:
Addressed both review comments in commit d110608:
1. Extracted `PROLOGUE_ID` as a class constant
2. Added `choice_from` and `choice_to` edges in the test

All CI checks pass (lint, type check, all Python versions). Ready to merge.

Issue #703, opened 2026-02-08
feat(fill): soft voice steering toward corpus-covered attribute combos


pvliesdonk:
## Problem
When using `corpus_only` exemplar strategy, the voice phase may select POV/tense/register combinations with no corpus coverage (e.g., third_omniscient, formal register). Small models then get no exemplar anchoring at all.

## Proposal
- Query `Corpus.list_exemplar_tags()` before voice phase
- Inject coverage hints into `fill_phase0_voice.yaml` as a preference guide (not hard constraint)
- Only inject when `exemplar_strategy` resolves to `corpus_only`
- Soft steering: "these combos are preferred" not "you must use these"

## Current Corpus Coverage
- POV: first, second, third_limited (missing: third_omniscient)
- Tense: past, present
- Register: conversational, literary, ornate, sparse (missing: formal)
- Rhythm: flowing, measured, punchy (missing: varied)
- Language: en only

## Acceptance Criteria
- [ ] Corpus coverage query before voice phase
- [ ] Coverage hints injected into voice prompt template
- [ ] Only when strategy is corpus_only
- [ ] Tests for coverage injection

Depends on: exemplar phase behavior issue
Deferred: implement after observing corpus_only results without steering

pvliesdonk:
Closing: voice steering moot — removing exemplars entirely.

Issue #702, opened 2026-02-08
feat(fill): make exemplar phase strategy-aware (skip LLM for small models)


pvliesdonk:
## Problem
When `exemplar_strategy` resolves to `corpus_only`, the exemplar phase should skip LLM fallback entirely rather than generating contaminating exemplars.

## Proposal
- Add `_resolve_exemplar_strategy()` method to FillStage
- Modify `_phase_0b_exemplar()`: when strategy is `corpus_only` and corpus has < 2 matches, skip exemplars entirely (no LLM call)
- Add `_log_corpus_coverage_suggestions()` to log nearby corpus combos that have coverage
- Voice document rules remain the primary style guide when no exemplars are available

## Acceptance Criteria
- [ ] `_resolve_exemplar_strategy()` resolves from config or auto-detects from tier
- [ ] `corpus_only` strategy skips LLM fallback, logs gap
- [ ] `full` strategy retains current LLM fallback behavior
- [ ] Coverage suggestions logged for nearby voice combos
- [ ] Tests in `test_fill_exemplars.py` covering all strategy paths

Depends on: exemplar strategy config issue
Part of: model-size-aware exemplar architecture (PRs 1-3)

pvliesdonk:
Closing: removing exemplars entirely instead of making them strategy-aware.

Issue #701, opened 2026-02-08
feat(fill): add exemplar_strategy config for model-aware exemplar selection


pvliesdonk:
## Problem
The exemplar phase always falls back to LLM generation when corpus has insufficient matches. For small models, this produces contaminated prose. Need a config-driven way to control exemplar behavior based on model capability.

## Proposal
- Add `exemplar_strategy: str = "auto"` to `FillConfig` (values: `auto`, `corpus_only`, `full`)
- `auto` (default): resolve from model capability tier (small → corpus_only, large → full)
- `corpus_only`: corpus exemplars only, no LLM fallback
- `full`: current behavior (corpus-first, LLM-fallback)
- Precedence: CLI flag > env var (`QF_EXEMPLAR_STRATEGY`) > project config > auto
- Pass creative model name through orchestrator → FillStage for auto-detection

## Acceptance Criteria
- [ ] `FillConfig.exemplar_strategy` field with `from_dict` support
- [ ] CLI flag `--exemplar-strategy`
- [ ] Env var `QF_EXEMPLAR_STRATEGY` support
- [ ] Model name passed from orchestrator to FillStage
- [ ] Unit tests for config resolution and precedence

Depends on: model capability tier issue
Part of: model-size-aware exemplar architecture (PRs 1-3)

pvliesdonk:
Closing: exemplar_strategy config no longer needed — removing exemplars entirely.

Issue #700, opened 2026-02-08
feat(providers): add model capability tier to ModelProperties


pvliesdonk:
## Problem
The FILL exemplar phase uses LLM-generated fallback exemplars when corpus has no match. Small models (<=8B params like qwen3:4b) copy exemplar content verbatim instead of abstracting style, contaminating prose with exemplar imagery. Need a way to classify models as "small" vs "large" to gate features appropriately.

## Proposal
- Add `capability_tier: str = "small"` field to `ModelProperties` in `model_info.py`
- Annotate all entries in `KNOWN_MODELS` (ollama 4b-8b: small, cloud providers: large)
- Add `get_capability_tier(provider, model) -> str` helper with name-pattern heuristic for unknown models
- Conservative default: unknown models → "small" (never risk exemplar contamination)

## Acceptance Criteria
- [ ] `ModelProperties` has `capability_tier` field
- [ ] All `KNOWN_MODELS` entries annotated
- [ ] `get_capability_tier()` function with heuristic fallback
- [ ] Unit tests for tier resolution (known models, unknown models, heuristic patterns)

Related: #695, #696
Part of: model-size-aware exemplar architecture (PRs 1-3)

pvliesdonk:
Closing: capability tier approach rejected — removing exemplars entirely instead.

PR #704, opened 2026-02-08
feat(providers): add model capability tier to ModelProperties (#700)

pvliesdonk:
## Problem
The FILL exemplar phase falls back to LLM-generated exemplars when the corpus has no match. Small models (<=8B params like qwen3:4b) copy exemplar *content* verbatim instead of abstracting *style*, contaminating prose with exemplar imagery. Test data showed LLM-generated exemplars leaked into 36-39 of 67 passages, worsening TTR (0.1015 → 0.0957).

We need a way to classify models as "small" vs "large" to gate features appropriately.

## Changes
- Add `capability_tier: str = "small"` field to `ModelProperties` dataclass
- Annotate all entries in `KNOWN_MODELS` registry (ollama 4b-8b: small, all cloud providers: large)
- Add `get_capability_tier(provider, model) -> str` helper function with:
  - KNOWN_MODELS lookup (highest priority)
  - Name-pattern heuristic for unknown models (detects `4b`, `7b`, `8b`, `13b` patterns)
  - Cloud provider default (large) for OpenAI/Anthropic/Google
  - Conservative default (small) for unknown local models
- Add 17 unit tests covering tier resolution for all paths

## Not Included / Future PRs
- Exemplar strategy config plumbing (#701)
- Strategy-aware exemplar phase behavior (#702)
- Voice steering toward corpus-covered combos (#703)

## Test Plan
```
uv run mypy src/questfoundry/providers/model_info.py  # Success
uv run ruff check src/questfoundry/providers/model_info.py tests/unit/test_model_info.py  # All passed
uv run pytest tests/unit/test_model_info.py -x -q  # 26 passed
```

## Risk / Rollback
- No behavior changes — this PR only adds a new field and helper function
- Default `capability_tier="small"` is backwards-compatible (existing code never reads it)
- Safe to revert by removing the field

### DIFF
--- a/src/questfoundry/providers/model_info.py
+++ b/src/questfoundry/providers/model_info.py
@@ -30,11 +30,20 @@ class ModelProperties:
     """Known properties for a specific model.
 
     Used in KNOWN_MODELS registry to consolidate model metadata.
+
+    Attributes:
+        context_window: Maximum input tokens the model can process.
+        supports_vision: Whether the model can process images.
+        supports_tools: Whether the model supports tool/function calling.
+        capability_tier: Model capability classification ("small" or "large").
+            Small models (<=13B params) copy exemplar content verbatim;
+            large models can abstract style from content.
     """
 
     context_window: int
     supports_vision: bool = False
     supports_tools: bool = True  # Most models support tools; o1 family doesn't
+    capability_tier: str = "small"  # "small" or "large"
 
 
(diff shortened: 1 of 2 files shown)


pvliesdonk:
Closing: decided to remove exemplars entirely rather than add model capability detection. See discussion in #706.

PR #710, opened 2026-02-08
feat(dress): batch Phase 1 briefs (5 per call)

pvliesdonk:
## Problem
DRESS Phase 1 makes one LLM call per passage, repeating ~900 tokens of
identical art direction + instructions each time. For a 70-passage story,
this is ~105k input tokens and 70 round-trips.

## Changes
- Add `format_passages_batch_for_briefs()` and `format_all_entity_visuals()`
  batch context helpers to `dress_context.py`
- Rewrite `_phase_1_briefs()` to batch 5 passages per LLM call using
  `BatchedBriefOutput` schema and `dress_brief_batch` template
- Uses `batch_llm_calls()` for bounded concurrency
- Composition log accumulates across batches for anti-repetition
- Remove unused imports (`format_passage_for_brief`, `format_entity_visuals_for_passage`,
  `DressPhase1Output`)
- Update existing Phase 1 tests to use `BatchedBriefOutput`
- Add test for batch grouping (7 passages -> 2 batches)
- Add 5 context formatter tests

**Depends on PR #709** (batch models + templates)

## Not Included / Future PRs
- Phase 2 codex batching (PR 4, independent)
- Individual item retry fallback (deferred — templates retained for this)

## Test Plan
- `uv run pytest tests/unit/test_dress_stage.py tests/unit/test_dress_context.py -x -q` — 93 passed
- `uv run mypy src/questfoundry/pipeline/stages/dress.py src/questfoundry/graph/dress_context.py` — no issues

## Risk / Rollback
- Changes Phase 1 behavior from per-passage to batched calls
- Old per-passage template (`dress_brief.yaml`) retained for future fallback
- Batch size of 5 is conservative (~1k output tokens, well within small model limits)

## Review Guide
Suggested order: `dress_context.py` (new helpers) -> `dress.py` (batching logic) -> tests

### DIFF
--- a/src/questfoundry/graph/dress_context.py
+++ b/src/questfoundry/graph/dress_context.py
@@ -279,3 +279,66 @@ def format_entity_visuals_for_passage(graph: Graph, passage_id: str) -> str:
                 lines.append(f"- **{raw_eid}**: {fragment}")
 
     return "\n".join(lines) if lines else ""
+
+
+def format_passages_batch_for_briefs(
+    graph: Graph,
+    passage_ids: list[str],
+    base_scores: dict[str, int],
+) -> str:
+    """Format a batch of passages for illustration brief generation.
+
+    Combines per-passage context (prose, metadata) with priority context
+    into a single markdown string for batched brief prompts.
+
+    Args:
+        graph: Graph containing passage, beat, and entity nodes.
+        passage_ids: Passage node IDs in this batch.
+        base_scores: Pre-computed structural scores keyed by passage_id.
+
+    Returns:
+        Formatted batch string with sections per passage.
+    """
+    from questfoundry.pipeline.stages.dress import describe_priority_context
+
+    sections: list[str] = []
+    for pid in passage_ids:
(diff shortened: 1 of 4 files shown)


pvliesdonk:
Addressed all review feedback:

**1. Circular import (Critical → Fixed)**: Moved `describe_priority_context` from `dress.py` to `dress_context.py`, eliminating the runtime circular import entirely.

**2. Silent batch failure (Valid → Fixed)**: Added `log.warning("brief_batch_failed", ...)` when `batch_result is None`.

**3. Race condition in composition log (By design)**: The composition log is only *read* during concurrent batch execution and only *mutated* in the sequential post-processing loop after `batch_llm_calls` returns. There's no actual race — concurrent batches see a stale snapshot, which is intentionally best-effort (documented in the plan). Added a clarifying comment.

**4. Test mock fragility (Valid → Fixed)**: Replaced `call_count`-based index with parsing passage IDs from batch context text using `f"### {pid}\n"` matching.

**5. Composition dedup**: Deferred — the anti-repetition log is already capped at 5 entries and duplicates are uncommon in practice. If it becomes an issue, easy to add.

PR #726, opened 2026-02-09
fix(orchestrator): restore pre-stage snapshot on re-run to prevent NodeExistsError

pvliesdonk:
## Problem

Re-running BRAINSTORM or SEED crashed with `NodeExistsError` because the orchestrator's mutation block unconditionally overwrote the clean pre-stage snapshot with the current (post-stage) graph, then tried to `create_node()` on nodes that already existed.

```
NodeExistsError: Node 'location::loc_kitchen_and_pantry' already exists
```

## Changes

- Add `_MUTATION_STAGE_PREREQUISITES` map to orchestrator (dream→None, brainstorm→dream, seed→brainstorm)
- Replace the unconditional `save_snapshot` with re-run detection: check `graph.get_last_stage()` against the stage's prerequisite
  - **First run** (`last_stage == prerequisite`): save clean pre-stage snapshot
  - **Re-run** (`last_stage != prerequisite`): restore from existing snapshot
  - **Fallback** (snapshot deleted): save current state with warning
- Add 6 tests covering first run, re-run, downstream cascade, fallback, and dream with None prerequisite

This matches the existing GROW/FILL/DRESS pattern (`grow.py:338-365`).

## Not Included / Future PRs

- DRESS is already safe (uses `upsert_node` throughout + internal snapshot management)
- DREAM is already safe (uses `upsert_node`)

## Test Plan

- `uv run ruff check` — all passed
- `uv run mypy src/questfoundry/pipeline/orchestrator.py` — no issues
- `uv run pytest tests/unit/test_orchestrator.py -x -q` — 50 passed (6 new)
- Pre-commit hooks all passed

## Risk / Rollback

- Low risk: only affects the mutation block, which runs after successful LLM generation
- The `last_stage` comparison approach naturally handles downstream cascades (when BRAINSTORM re-runs, SEED sees its prerequisite and treats next run as first run)

### DIFF
--- a/src/questfoundry/pipeline/orchestrator.py
+++ b/src/questfoundry/pipeline/orchestrator.py
@@ -55,6 +55,46 @@ _MUTATION_STAGE_PREREQUISITES: dict[str, str | None] = {
 }
 
 
+def _load_graph_for_mutation(project_path: Path, stage_name: str) -> Graph:
+    """Load graph with re-run detection for orchestrator-managed mutation stages.
+
+    On first run (last_stage == prerequisite), saves a clean pre-stage snapshot.
+    On re-run, restores from the snapshot to avoid NodeExistsError from
+    create_node() on nodes that already exist.
+
+    Raises:
+        ValueError: If re-running but the pre-stage snapshot is missing.
+    """
+    graph = Graph.load(project_path)
+    last_stage = graph.get_last_stage()
+    prerequisite = _MUTATION_STAGE_PREREQUISITES.get(stage_name)
+    snapshot_path = project_path / "snapshots" / f"pre-{stage_name}.json"
+
+    if last_stage == prerequisite:
+        # First run: save clean pre-stage snapshot
+        save_snapshot(graph, project_path, stage_name)
+    elif snapshot_path.exists():
+        # Re-run: restore clean pre-stage graph
+        graph = Graph.load_from_file(snapshot_path)
+        log.info(
+            "rerun_restored_snapshot",
+            stage=stage_name,
+            from_last_stage=last_stage,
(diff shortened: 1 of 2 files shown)


pvliesdonk:
**Review responses (all addressed in 62b57e9):**

**@claude[bot] re: fallback saves corrupted snapshot (line 704):** Fixed. Changed the fallback to raise `ValueError` with remediation guidance, matching the GROW pattern (grow.py:347-351). Silent corruption is now impossible.

**@claude[bot] re: missing 'seed' entry (line 51):** The `seed` entry is already present — lines 51-55 show all three entries: `dream: None`, `brainstorm: dream`, `seed: brainstorm`. The reviewer appears to have missed it.

**@claude[bot] re: test doesn't verify fallback behavior (line 1187):** Fixed. Test renamed to `test_rerun_no_snapshot_raises_error` and now asserts `pytest.raises(ValueError, match="requires the pre-stage snapshot")`.

**@claude[bot] re: downstream cascade test doesn't test SEED (line 1153):** The test verifies the prerequisite condition (`last_stage == prerequisite`) which is the mechanism that enables the cascade. Running actual SEED mutations would require valid brainstorm+seed artifact data and test the mutation handler (already covered by mutation tests), not the re-run detection logic.

**@gemini-code-assist[bot] re: extract helper method (line 712):** Done. Extracted `_load_graph_for_mutation()` as a module-level function. The mutation block now calls it with a single line: `graph = _load_graph_for_mutation(self.project_path, stage_name)`.

**@gemini-code-assist[bot] re: test helper duplicates production code (line 1064):** Fixed. `_simulate_mutation_block()` now calls the production `_load_graph_for_mutation()` directly instead of duplicating the logic.

PR #732, opened 2026-02-09
fix(seed): silently deduplicate identical LLM entries instead of retrying

pvliesdonk:
## Problem

When the LLM emits identical duplicate entries during SEED serialization, the inner
retry loop asks it to fix them. The LLM "fixes" by renaming (e.g.,
`object_cctv_archive` → `object_cctv_archive_duplicate_check`), creating phantom IDs
that fail semantic validation, triggering the expensive outer retry loop. Investigation
confirmed these are true identical duplicates (not name clashes or brainstorm cascading).

## Changes

- `_check_ids_unique()` → `_deduplicate_and_check()`: silently drops identical copies
  (same ID + equal model content), only raises on non-identical duplicates (same ID,
  different content — a real conflict)
- Error message for non-identical duplicates says "conflicting content" and instructs
  removal (not renaming)
- `PathBeatsSection` consolidated to use the shared helper instead of inline duplicate logic
- Tests updated: identical duplicates are accepted (deduplicated), non-identical rejected

## Not Included / Future PRs

- No mutation-layer changes (Pydantic-level dedup prevents the issue before mutations run)

## Test Plan

- `uv run pytest tests/unit/test_seed_models.py -x -q` — 21 tests pass
- `uv run pytest tests/unit/test_serialize.py -x -q` — 59 tests pass
- `uv run ruff check src/questfoundry/models/seed.py` — clean
- `uv run mypy src/questfoundry/models/seed.py` — clean

## Risk / Rollback

Low risk. Only changes Pydantic validation behavior for section wrappers.
Identical duplicates that previously caused retry loops are now silently handled.
Non-identical duplicates still produce actionable error messages.


pvliesdonk:
Addressed gemini's review: improved error message phrasing to `Duplicates found for {type_name}: ...` — avoids the awkward `entity_ids` pluralization. Pushed in 33302f7.

PR #733, opened 2026-02-09
fix(grow): Phase 8c overlay schema incompatible with OpenAI strict mode

pvliesdonk:
## Problem

GROW Phase 8c crashes on OpenAI models (gpt-5-mini) with:

```
Invalid schema for response_format 'Phase8cOutput': Extra required key 'details' supplied.
```

`OverlayProposal.details: dict[str, str]` generates an `additionalProperties` schema,
which OpenAI strict structured output mode forbids. All object schemas must have
explicit `properties` with `additionalProperties: false`.

## Changes

- Add `OverlayDetailItem` model with `key` and `value` fields
- Change `OverlayProposal.details` from `dict[str, str]` to `list[OverlayDetailItem]`
- Add `details_as_dict()` helper method for graph storage conversion
- Update `grow.py` to call `details_as_dict()` when storing overlays
- Update prompt template to show the `[{key, value}]` format
- `EntityOverlay` (graph-side model) unchanged — not used with structured output

## Not Included / Future PRs

- No audit of other models for `dict[str, str]` fields (only these two instances exist)

## Test Plan

- `uv run pytest tests/unit/test_grow_models.py::TestOverlayProposal -x -q` — 5 pass
- `uv run pytest tests/unit/test_grow_validators.py::TestValidatePhase8cOutput -x -q` — 4 pass
- `uv run pytest tests/unit/test_grow_stage.py::TestPhase8cOverlays -x -q` — 4 pass
- Schema verification: `_make_all_required(Phase8cOutput.model_json_schema())` produces
  valid schema with `additionalProperties: false` throughout

## Risk / Rollback

Low. Changes LLM output format for Phase 8c only. Graph storage format unchanged
(still stored as `dict[str, str]` via `details_as_dict()` conversion).


pvliesdonk:
Addressed both reviews in 4cadca9:

**gemini** — Added `match="at least 1"` to `test_empty_details_rejected` for specificity.

**claude** — Added `_check_no_duplicate_keys` model_validator to `OverlayProposal`. Duplicate keys now raise `ValueError("Duplicate keys in details: ...")` instead of silently dropping values in `details_as_dict()`. Added test `test_duplicate_keys_rejected`.

PR #822, opened 2026-02-12
refactor(grow): per-dilemma convergence point selection

### LINKED ISSUE #821: refactor(grow): per-dilemma convergence point selection
## Problem

After fixing #814 (single-explored dilemma filtering), convergence funneling persists in seeds where multiple divergent dilemmas share `hard` convergence policy. All arcs converge at the same beat because convergence is computed per-arc, and `hard` dominates across all dilemma-arc pairs.

Evidence from test-big-hard-2 seed (runs 2b and 2c):
- 8 dilemmas, 4 divergent (3 hard, 1 soft)
- All 8 non-spine arcs converge at the same beat (`beat::watcher_appear_visible_or_disappear_fully__disappear_fully_beat_01`)
- #814 fix correctly filters single-explored dilemmas, but hard still dominates

## Desired Behavior

Each dilemma should resolve its convergence independently. Two arcs that differ only on dilemma A should converge at dilemma A's convergence point, not at a global convergence point driven by dilemma B's hard policy.

## Proposed Approach

Refactor `find_convergence_points()` to compute convergence per-dilemma rather than per-arc:

1. For each divergent dilemma, find the beat where its paths rejoin (independent of other dilemmas)
2. Each arc inherits convergence points from its constituent dilemmas
3. The `hard`/`soft`/`flavor` policy applies within a single dilemma's convergence, not globally across all dilemmas in an arc

This is a non-trivial algorithm change — the current arc-level approach treats the Cartesian product of paths as the unit of convergence, but dilemma-level convergence is the narratively correct granularity.

## Context

- #814 fixed single-explored dilemma filtering (PR #818)
- Current algorithm: `_find_arc_dilemma_policies()` → `_get_effective_policy()` → `_find_convergence_candidates()` in `grow_algorithms.py`
- Test seed: test-big-hard-2 (8 dilemmas, 16 arcs)

pvliesdonk:
## Problem

Convergence funneling: all branch arcs converge at the same beat because `find_convergence_points()` computes ONE convergence point per arc by combining all dilemma policies into a single effective policy. Arcs differing on different dilemmas get the same convergence point.

Closes #821

## Changes

- Add `DilemmaPolicy` and `DilemmaConvergence` dataclasses in `grow_algorithms.py`
- Refactor `_find_arc_dilemma_policies()` to return `list[DilemmaPolicy]` instead of `list[tuple[str, int]]`
- Update `_get_effective_policy()` to consume `DilemmaPolicy` objects
- Add `_compute_per_dilemma_convergence()` helper that computes convergence separately for each dilemma (filtering beats to that dilemma + neutral)
- Simplify `find_convergence_points()` main loop: delegate to per-dilemma computation, pick earliest non-None as arc-level result
- Add `dilemma_convergences` field to `ConvergenceInfo` dataclass
- Phase 7 in `grow.py` writes `dilemma_convergences` list to arc nodes
- Update 2 Phase 7 integration tests to set explicit budget (per-dilemma budget applies to each dilemma's own exclusive beats)
- Add 4 new tests for per-dilemma convergence behavior

## Not Included / Future PRs

- FILL context enrichment using per-dilemma convergence (#821 follow-up)
- Inspection report per-dilemma display
- Validation cross-referencing

## Test Plan

```bash
uv run mypy src/questfoundry/graph/grow_algorithms.py src/questfoundry/pipeline/stages/grow.py  # Clean
uv run ruff check src/  # Clean
uv run pytest tests/unit/test_grow_algorithms.py -x -q  # 207 passed
```

- All 12 existing `TestFindConvergencePointsPolicyAware` tests pass (backward compat)
- All 6 `TestFindConvergencePoints` tests pass
- All 4 `TestPhase7Integration` tests pass
- 4 new per-dilemma tests pass

## Risk / Rollback

- Per-dilemma budget is checked against each dilemma's own exclusive beats (not the combined set). This is a behavior change: arcs where each dilemma has fewer exclusive beats than the budget will no longer converge. This is the correct behavior per spec (budget means "beats needed for this dilemma's payoff").
- Arc-level `converges_at` is the earliest non-None per-dilemma convergence — backward compatible for single-dilemma arcs.

## Review Guide

1. Start with `DilemmaPolicy` and `DilemmaConvergence` dataclasses (top of convergence section)
2. `_find_arc_dilemma_policies()` return type change
3. `_compute_per_dilemma_convergence()` — the core new logic
4. `find_convergence_points()` main loop simplification
5. Phase 7 graph writing in `grow.py`
6. New tests at end of `TestFindConvergencePointsPolicyAware`

### DIFF
--- a/src/questfoundry/graph/grow_algorithms.py
+++ b/src/questfoundry/graph/grow_algorithms.py
@@ -1300,11 +1300,7 @@ def _compute_per_dilemma_convergence(
             converges_at = _find_convergence_for_soft(filtered_branch, filtered_spine, dp.budget)
         else:
             # flavor: first shared beat
-            converges_at = None
-            for beat_id in filtered_branch:
-                if beat_id in filtered_spine:
-                    converges_at = beat_id
-                    break
+            converges_at = next((b for b in filtered_branch if b in filtered_spine), None)
 
         results.append(
             DilemmaConvergence(
@@ -1413,14 +1409,12 @@ def find_convergence_points(
             # Arc-level converges_at = earliest non-None per-dilemma convergence
             non_none = [dc for dc in per_dilemma if dc.converges_at]
             if non_none:
-                converges_at = min(
-                    non_none,
-                    key=lambda dc: (
-                        branch_after_div.index(dc.converges_at)
-                        if dc.converges_at in branch_after_div
-                        else len(branch_after_div)
-                    ),
-                ).converges_at
+
+                def _beat_index(dc: DilemmaConvergence, seq: list[str] = branch_after_div) -> int:
+                    at = dc.converges_at
+                    return seq.index(at) if at and at in seq else len(seq)
+
+                converges_at = min(non_none, key=_beat_index).converges_at
 
         result[arc.arc_id] = ConvergenceInfo(
             arc_id=arc.arc_id,
--- a/src/questfoundry/graph/grow_algorithms.py
+++ b/src/questfoundry/graph/grow_algorithms.py
@@ -1025,6 +1025,40 @@ def compute_divergence_points(
 
 
 @dataclass
+class DilemmaPolicy:
+    """Policy info for a single divergent dilemma within an arc.
+
+    Attributes:
+        dilemma_id: Scoped dilemma ID (e.g. ``dilemma::foo``).
+        policy: Convergence policy (hard/soft/flavor).
+        budget: Payoff budget for this dilemma.
+        non_canon_path_id: The non-canonical path for this dilemma in this arc.
+    """
+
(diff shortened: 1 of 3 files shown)


pvliesdonk:
Addressed both review comments:

- **@claude[bot]** (type safety in `min()` lambda): Extracted to a named `_beat_index()` function with default-arg binding — avoids the closure-over-loop-variable issue (ruff B023) and makes type narrowing explicit.
- **@gemini-code-assist** (`next()` for flavor loop): Replaced the `for`/`break` loop with `next((b for b in filtered_branch if b in filtered_spine), None)`.

PR #832, opened 2026-02-12
feat(viz): add overlay and grants styling to graph visualization

pvliesdonk:
## Problem
The `qf graph` command showed basic passage/choice structure but didn't visually distinguish state-changing elements: passages with codeword-dependent prose (overlays) looked identical to regular passages, and choices that grant codewords were indistinguishable from plain choices.

## Changes
- **Overlay passages** (containing entities with codeword-dependent content) get a thick orange-red border in DOT output and dedicated CSS classes (`overlay`, `startOverlay`, `endingOverlay`) in Mermaid
- **Grants edges** (choices that set codewords) rendered in slate blue (`#6A5ACD`) with thick lines in both DOT and Mermaid
- `VizNode` gains `has_overlays` (bool) and `outgoing_count` (int) fields
- `VizEdge` gains `grants` (list[str]) field
- `build_story_graph` detects overlay-affected passages by tracing entity overlays through both raw and scoped ID formats
- 16 new tests covering overlay detection, grants edge rendering, and outgoing count calculation

## Not Included / Future PRs
- Visual distinction for linear vs branching passages (data is available via `outgoing_count` but no rendering difference yet)
- Legend/key in output explaining color coding

## Test Plan
```bash
uv run pytest tests/unit/test_visualization.py -x -q  # 37 passed
uv run mypy src/questfoundry/visualization.py          # no issues
uv run ruff check src/questfoundry/visualization.py    # all passed
uv run qf graph -p test-murder-gpt5mini-retry -f dot   # verified overlay borders + grants colors
uv run qf graph -p test-murder-gpt5mini-retry -f json  # verified new fields in JSON output
```

Test project stats: 55/74 overlay passages, 17/97 grants edges, 0 gated choices.

## Risk / Rollback
- Additive change only — no existing behavior modified
- JSON output gains new fields (`has_overlays`, `outgoing_count`, `grants`) which is backward compatible

### DIFF
--- a/src/questfoundry/visualization.py
+++ b/src/questfoundry/visualization.py
@@ -118,7 +118,7 @@ def build_story_graph(
     # Track which passages contain overlay-affected entities
     overlay_passages: set[str] = set()
     for pid, pdata in passages.items():
-        for ent in pdata.get("entities", []):
+        for ent in pdata.get("entities") or []:
             # Check both raw ID and scoped forms
             if ent in overlay_entity_ids:
                 overlay_passages.add(pid)
--- a/src/questfoundry/visualization.py
+++ b/src/questfoundry/visualization.py
@@ -30,6 +30,8 @@ _ARC_COLORS = [
 _SHARED_COLOR = "#D3D3D3"  # light grey for multi-arc passages
 _START_COLOR = "#90EE90"  # light green
 _ENDING_COLOR = "#FFB6C1"  # light pink
+_GRANTS_COLOR = "#6A5ACD"  # slate blue for state-changing choices
+_OVERLAY_BORDER = "#FF4500"  # orange-red border for overlay passages
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


pvliesdonk:
Addressed all review comments in af08901 and 1a3a5d3:

1. **Nested loop break bug** — Replaced inner `for prefix in ...` loop with `any()` so `break` correctly exits the outer entity loop.
2. **Requires/grants precedence** — Added comment explaining `elif` logic; requires (gated) takes precedence over grants (state-changing). Added test `test_requires_takes_precedence_over_grants`.
3. **DOT/Mermaid is_return inconsistency** — DOT now also skips grants color for return edges (`not edge.is_return`), matching Mermaid behavior. Added test `test_return_edge_with_grants_keeps_dashed_style`.
4. **Defensive entities=None** — Changed `.get("entities", [])` to `.get("entities") or []` to handle explicit `None`. Added test `test_entities_none_does_not_crash`.
5. **Raw-ID overlay detection** — Added test `test_overlay_detected_via_raw_id` confirming the `any()` prefix matching works correctly.
6. **Extracted `_ENTITY_PREFIXES`** — Constant at module level for overlay prefix matching.

PR #835, opened 2026-02-12
fix(prompts): add anti-helpfulness guards to SEED and DREAM discuss prompts

pvliesdonk:
## Problem
GPT-5 was offering follow-up work during interactive seed sessions ("If you want, next I can: turn this into a flow diagram, expand paths into scene flows, present retain/cut as a checklist..."). PR #718/#720 added anti-helpfulness guards to BRAINSTORM but missed SEED and DREAM.

## Changes
- **`discuss_seed.yaml`**: Add 3 "Do NOT" items matching brainstorm pattern (no follow-up offers, no previewing later stages, no chat-style endings), STOP instruction in non-interactive section, scope reminder in interactive section
- **`discuss.yaml`**: Add STOP instruction in non-interactive section

## Not Included
- FILL and DRESS discuss templates already have adequate guards

## Test Plan
- Prompt-only change; no code or tests affected
- Verified YAML syntax passes `check yaml` pre-commit hook

## Risk / Rollback
- Zero risk — additive prompt constraints only
- No code changes

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

PR #836, opened 2026-02-12
refactor(prompts): replace anti-helpfulness blacklists with task sandbox guardrails

pvliesdonk:
## Problem

GPT-5's RLHF training causes it to offer follow-up work after completing stage tasks (flow diagrams, scene expansions, checklists). PR #835 attempted to fix this with "Do NOT offer follow-up" blacklists, but this fights the model's training — the model treats prohibitions as suppressible suggestions. Reframing as **task sandboxing** leverages the training instead: define what the task IS, explain WHY out-of-scope work is discarded.

Replaces (closed) PR #835.

## Changes

- **New shared component** `prompts/components/task_sandbox.yaml` — single source of truth for task boundary enforcement, explains that the pipeline discards out-of-scope work
- **Task Contract** added to all 5 discuss templates (DREAM, BRAINSTORM, SEED, FILL, DRESS) — front-loads stage position, deliverables, and completion condition
- **Sandbox injection** via `{sandbox_section}` placeholder in all templates, loaded by `load_sandbox_section()` in `agents/prompts.py`
- **Format Pitfalls** — renamed from "What NOT to Do", retaining only format/validation-specific items (ID naming, prose vs notes)
- **Interactive redirect** — gives model behavior TO DO ("acknowledge interest, redirect back to scope") instead of prohibition
- **Simplified mode sections** — removed redundant STOP/follow-up instructions now handled by sandbox

### Design Principles (from prompt-engineer + llm-architect consultation)

1. **Positive framing** — "Your task is COMPLETE when X" instead of "Do NOT offer Y"
2. **Causal explanation** — "pipeline discards it" aligns with RLHF (model wants to be helpful → learns this IS helpful)
3. **One check, not eight** — model checks against contract once, not against N prohibitions
4. **Shared component** — consistency across all stages, single place to update

## Not Included / Future PRs

- Full `components` field / PromptCompiler component system (components: [] field exists but is never wired)
- System-level response trimming in `run_discuss_phase` (if data shows models still escape sandbox)
- Provider-aware sandwich repetition (if GPT-5 specifically needs end-of-prompt reinforcement)

## Test Plan

- `pre-commit run check-yaml` on all template files — passed
- `uv run mypy` on all 3 modified Python files — passed
- `uv run ruff check` on all 3 modified Python files — passed
- Import + rendering test for all 5 templates — passed (sandbox section injected, Task Contract present)
- No unit tests exist for prompt text content (prompts are tested at integration level)

## Risk / Rollback

- Low risk: prompt text changes only affect LLM behavior, not pipeline correctness
- Rollback: revert commit, re-add blacklists
- Templates validated by rendering test — no placeholder mismatches

### DIFF
--- a/prompts/components/task_sandbox.yaml
+++ b/prompts/components/task_sandbox.yaml
@@ -1,0 +1,20 @@
+name: task_sandbox
+description: Shared task boundary enforcement for all discuss phases
+
+content: |
+  ## Task Boundary
+
+  You are one stage in a multi-stage pipeline. Your deliverables are defined
+  above — everything outside them is handled by other stages with their own
+  prompts, context, and validation.
+
+  Work beyond your deliverables is discarded by the pipeline. Later stages
+  cannot see your conversation, only the structured artifact this stage
+  produces. Offering downstream work (diagrams, expansions, checklists,
+  scene flows) is wasted effort — the pipeline discards it.
+
+  **When your deliverables are complete, end your response.** Do not:
+  - Summarize what you produced
+  - Suggest next steps or offer follow-up work
+  - Preview what later stages will do
+  - End with "let me know if..." or "If you want, next I can..."
--- a/prompts/templates/discuss.yaml
+++ b/prompts/templates/discuss.yaml
@@ -3,7 +3,11 @@ description: Creative exploration phase for interactive fiction development
 
 system: |
   You are a creative collaborator establishing HIGH-LEVEL VISION for interactive fiction.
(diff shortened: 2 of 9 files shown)


pvliesdonk:
Re: inline imports — keeping them inline is intentional. Both `fill.py` and `dress.py` use inline imports extensively throughout (e.g., `from questfoundry.agents import run_discuss_phase` at line 801 in fill.py) to avoid circular import chains between `pipeline.stages.*` and `agents.*`. Consistent with existing pattern.

Merging — all checks green, review LGTM.

Issue #895, opened 2026-02-14
bug(grow): 3 passages have only gated forward choices (potential soft-lock)


pvliesdonk:
## Problem

`qf inspect` reports 3 passages where ALL forward choices require codewords:

```
passage::elias_reveal_water_grid_log_or_ignore_it__ignore_beat_03
passage::marcus_distribute_rations_fairly_or_falsify_records__distribute_fairly_beat_04
passage::marcus_distribute_rations_fairly_or_falsify_records__falsify_records_beat_04
```

If a player reaches one of these passages without the required codewords, they have no available forward choices — a **soft-lock**.

## Evidence

The `forward_path_reachability` check in `qf inspect` flags these. The 12 paradoxical gates (#890) contribute to this: players on non-spine paths encounter choices requiring spine codewords they can never have.

## Impact

Players on certain arc combinations may hit dead ends where no choice is available.

## Fix

Depends on #890 (paradoxical gates). Once gates are corrected to only require discriminating codewords, the soft-lock issue should resolve.

## Related

- #890 (paradoxical codeword gates — root cause)

pvliesdonk:
Resolved by #890 (PR #903). The 3 soft-locked passages were caused by paradoxical 4-codeword gates on non-ending choices. `compute_all_choice_requires()` was fixed to use branch-exclusive codewords (off-spine choices the player DID make) instead of spine-exclusive codewords (choices they did NOT make). With correct gates, all forward choices are satisfiable for the appropriate player arc.

PR #903, opened 2026-02-14
fix(grow): use branch-exclusive codewords for gate computation

### LINKED ISSUE #890: bug(grow): 12 non-ending choices have paradoxical 4-codeword requires gates
## Problem

The `test-new` audit reveals 12 non-ending choice nodes that require ALL 4 spine codewords:
```
requires: [clinic_access_granted, entry_granted, log_revealed, rations_distributed_fairly]
```

These guard transitions to **non-spine paths** (e.g., "Elias ignores the log", "Lina withdraws access"). A player on the spine path already made those choices and would never take these alternative routes. These gates are **logically unsatisfiable in context** — dead choices the player can never reach.

## Evidence

From `test-new` graph.db:
- 28/84 choices have non-empty `requires`
- 16 are ending choices (correct — unique 3-codeword combinations)
- 12 are non-ending choices (all require the same 4 spine codewords)

Example:
```
choice::prologue__cassia_remain_silent_beat_01
  requires: [clinic_access_granted, entry_granted, log_revealed, rations_distributed_fairly]
```
This choice leads to the "remain silent" path but requires the "allow entry" codeword — contradictory.

## Root Cause

`compute_all_choice_requires()` computes requirements based on arc membership, but doesn't distinguish between:
1. Gates that discriminate between arc families (correct)
2. Gates that require codewords from arcs the player is already NOT on (paradoxical)

## Impact

- 12 dead choices clutter the passage graph
- `qf inspect` passes (`gate_satisfiability` checks grantability, not reachability)
- Players on non-spine paths can never satisfy these gates

## Fix

`compute_all_choice_requires()` should only require codewords that are **discriminating** — codewords that distinguish between arcs that BOTH include the source passage. Codewords from arcs the source passage isn't in should be excluded.

pvliesdonk:
## Problem

12 non-ending choices in `test-new` require ALL 4 spine codewords — a set that is logically unsatisfiable because the player chose specific paths at each dilemma and cannot earn codewords from paths they didn't take. This also causes #895 (3 soft-locked passages).

Closes #890

## Root Cause

`compute_all_choice_requires()` collected codewords from **spine-exclusive** paths (paths the player did NOT take) instead of **branch-exclusive** paths (the off-spine choices the player DID make). Additionally, it used union across arcs instead of intersection, compounding the problem for multi-arc passages.

## Changes

- **`grow_algorithms.py`**: Fix `compute_all_choice_requires()` to collect from branch-exclusive paths and use intersection across arcs
- **`grow_validation.py`**: Add `check_gate_co_satisfiability()` — validates that all required codewords are co-reachable in a single playthrough (detects paradoxical gates). Wired into `validate_grow_output()`.
- **`test_grow_algorithms.py`**: Update 2 existing tests + add multi-arc intersection test
- **`test_grow_validation.py`**: Add 3 tests for co-satisfiability validation

## Not Included / Future PRs

- #895 (soft-locked passages) — should be resolved by this fix since the paradoxical gates were the root cause
- Re-running GROW on test-new to verify gate counts

## Test Plan

```bash
uv run mypy src/questfoundry/ && uv run ruff check src/  # clean
uv run pytest tests/unit/test_grow_algorithms.py tests/unit/test_grow_validation.py -x -q  # 321 passed
```

## Risk / Rollback

- The gate computation change affects which codewords are required for branch-exclusive passages
- Existing `check_gate_satisfiability` (global grantability) continues to work alongside the new `check_gate_co_satisfiability` (per-playthrough reachability)
- Rollback: revert the `compute_all_choice_requires()` change

## Review Guide

1. Start with `grow_algorithms.py` — the core algorithm fix (spine-exclusive → branch-exclusive, union → intersection)
2. Then `grow_validation.py` — the new co-satisfiability check
3. Then tests — verify the expected behavior matches the new algorithm

### DIFF
--- a/src/questfoundry/graph/grow_algorithms.py
+++ b/src/questfoundry/graph/grow_algorithms.py
@@ -1971,10 +1971,14 @@ def compute_all_choice_requires(
     """Compute codeword ``requires`` for hard-policy branch entry passages.
 
     For each passage that appears exclusively in hard-policy branch arcs
-    (not on the spine), collects codewords from **spine-exclusive** paths —
-    paths on the spine that are NOT shared with the branch arc.  These
-    codewords are earnable before the branch divergence point, making the
-    gate satisfiable in a single playthrough.
+    (not on the spine), collects codewords from **branch-exclusive** paths —
+    paths on the branch arc that are NOT on the spine.  These codewords
+    identify the specific off-spine choices the player made, so the gate
+    is satisfiable by any player on that arc.
+
+    When a passage appears on multiple hard arcs, the requires are the
+    **intersection** of each arc's branch-exclusive codewords — ensuring
+    any player on any qualifying arc can satisfy the gate.
 
     Returns:
         Mapping of ``passage_id`` → list of required codeword IDs.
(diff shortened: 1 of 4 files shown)


pvliesdonk:
Addressed both review comments in 9dbc77d:

1. **Gemini (line 2046)**: Using `set.intersection(*arc_codeword_sets)` for conciseness
2. **Claude (line 2040)**: Added clarifying comment explaining why empty arc codeword sets are skipped before intersection

PR #905, opened 2026-02-14
fix(grow): policy-aware arc limit counts only hard dilemmas

### LINKED ISSUE #904: GROW enumerate_arcs arc limit counts soft dilemmas toward ceiling
## Problem

`enumerate_arcs()` checks the total arc count (`2^all_explored_dilemmas`) against the limit, but soft/flavor dilemmas converge back and don't multiply endings. With 1 hard + 7 soft dilemmas, the total arc count is 2^8 = 256, which exceeds the 64-arc limit and fails GROW — even though only 2^1 = 2 ending-distinct arcs exist.

Observed on `projects/test-new2`: GROW fails at Phase 5 (enumerate_arcs) with `Arc count (256) exceeds limit of 64`.

## Root Cause

The arc count check in `enumerate_arcs()` (grow_algorithms.py:1224-1231) uses `len(arcs)` — the total Cartesian product — instead of counting only hard-policy dilemmas that actually multiply endings.

## Expected Behavior

The arc count limit should only count hard-policy dilemmas. Soft/flavor dilemmas converge and don't produce distinct endings, so they shouldn't trigger the limit.

pvliesdonk:
## Problem

`enumerate_arcs()` checks the total arc count (2^all_explored_dilemmas) against the 64-arc limit. Soft/flavor dilemmas converge back and don't multiply endings, but they still multiplied the arc count. With 1 hard + 7 soft dilemmas (test-new2), total arcs = 2^8 = 256, which exceeds the limit — even though only 2^1 = 2 ending-distinct arcs exist.

Fixes #904

## Changes

- **`src/questfoundry/graph/grow_algorithms.py`**: Arc count limit now only counts hard-policy dilemmas with 2+ paths. The full Cartesian product is still enumerated (downstream phases need soft-variant arcs for convergence metadata and residue), but the limit check uses effective arc count = 2^(hard dilemmas). Unclassified dilemmas default to soft.
- **`tests/unit/test_grow_algorithms.py`**: Updated `test_combinatorial_limit` to use hard dilemmas. Added `test_soft_dilemmas_do_not_count_toward_limit` (1 hard + 7 soft = 256 total arcs succeeds) and `test_unclassified_dilemmas_default_to_soft` (7 unclassified dilemmas = 128 arcs succeeds).

## Not Included / Future PRs

- N/A — this is a complete fix for the arc limit behavior

## Test Plan

```bash
uv run pytest tests/unit/test_grow_algorithms.py::TestEnumerateArcs -x -q  # 13 passed
uv run pytest tests/unit/test_grow_algorithms.py tests/unit/test_grow_validation.py -x -q  # 323 passed
uv run mypy src/questfoundry/graph/grow_algorithms.py  # clean
uv run ruff check src/questfoundry/graph/grow_algorithms.py  # clean
```

## Risk / Rollback

Low risk. The only change is in the arc count check — the Cartesian product and all downstream processing are unchanged. Worst case: more arcs are enumerated than before (up to 2^8 = 256 for 8 explored dilemmas), but all GROW phases are deterministic and handle this fine.

### DIFF
--- a/src/questfoundry/graph/grow_algorithms.py
+++ b/src/questfoundry/graph/grow_algorithms.py
@@ -1146,16 +1146,21 @@ def enumerate_arcs(graph: Graph, *, max_arc_count: int | None = None) -> list[Ar
     The spine arc contains all canonical paths. Branch arcs contain at least
     one non-canonical path.
 
+    The arc count limit is **policy-aware**: only hard-policy dilemmas count
+    toward the limit, since soft/flavor dilemmas converge and don't multiply
+    endings.  The full Cartesian product is still enumerated so downstream
+    phases (convergence metadata, residue) can operate on soft-variant arcs.
+
     Args:
         graph: Graph containing dilemma, path, and beat nodes.
-        max_arc_count: Safety ceiling for arc count. Defaults to
-            ``_MAX_ARC_COUNT`` (64) if not provided.
+        max_arc_count: Safety ceiling for effective (hard-only) arc count.
+            Defaults to ``_MAX_ARC_COUNT`` (64) if not provided.
 
     Returns:
         List of Arc models, spine first, then branches sorted by ID.
 
     Raises:
-        ValueError: If arc count exceeds the limit.
+        ValueError: If effective arc count exceeds the limit.
     """
     dilemma_nodes = graph.get_nodes_by_type("dilemma")
     path_nodes = graph.get_nodes_by_type("path")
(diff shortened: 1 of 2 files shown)


pvliesdonk:
Addressed Gemini's suggestion in 9b3d96d: used `sum()` generator expression and dropped the unnecessary `2**0` guard since `2**0 == 1` is already correct.

PR #907, opened 2026-02-14
fix(grow): dilemma-aware round-robin beat interleaving

### LINKED ISSUE #893: bug(grow): commits_timing clusters early or late instead of mid-arc distribution
## Problem

The `test-new` audit shows commit timing warnings from `qf inspect`:

- Cassia `allow_entry`: commits at position **2/20** (too early, <3 beats of setup)
- Cassia `remain_silent`: commits at position **3/19** (too early)
- Marcus `distribute_fairly`: commits at position **19/20** (too late, >80%)
- Marcus `falsify_records`: commits at position **20/20** (too late, >80%)

Commits should land between 20-80% of the arc, with enough setup beats before and payoff beats after.

## Root Cause

SEED beat ordering places each dilemma's beats in a fixed sequence. The first dilemma (Cassia) commits early because it's at the start of the story. The last dilemma (Marcus) commits late because it's near the end. The commit beat position is determined by SEED's beat ordering, not by any GROW-time adjustment.

## Impact

- Early commits mean the player makes irrevocable choices before understanding consequences
- Late commits mean there's no payoff — the ending comes immediately after the choice
- This is the existing #840 issue confirmed with real data

## Related

- #840 (same issue, previously theoretical)

pvliesdonk:
## Problem

Commit beats (dilemma resolution) cluster at the beginning or end of arc sequences instead of being distributed. In test-new, Cassia's commits is at position 2/20 (10%) and Marcus's at 20/20 (100%).

`topological_sort_beats()` uses alphabetical tie-breaking when multiple beats have in-degree 0. Since there are no `requires` edges between beats from different dilemmas, all of dilemma A's beats sort before dilemma B's beats purely by ID order.

Closes #893

## Changes

- **`grow_algorithms.py`**: Add dilemma-aware round-robin to `topological_sort_beats()`:
  - Build `beat → dilemma` mapping from `dilemma_impacts` metadata
  - Track per-dilemma emission count during Kahn's algorithm
  - Sort key changes from `(priority_tier, beat_id)` to `(priority_tier, emission_count, beat_id)`
  - Re-sorts queue after each emission to reflect updated counts
- **`test_grow_algorithms.py`**: 5 new tests covering interleaving, requires override, priority precedence, no-impacts fallback

## Not Included / Future PRs

- Changes to `check_commits_timing()` thresholds (the validation correctly detects the problem; the sort was the root cause)

## Test Plan

```bash
uv run pytest tests/unit/test_grow_algorithms.py -x -q -k "TestTopologicalSort"
# 15 passed
uv run pytest tests/unit/test_grow_algorithms.py -x -q
# 244 passed (no regressions)
uv run mypy src/questfoundry/graph/grow_algorithms.py
# Success
```

## Risk / Rollback

- Low risk: the sort key is extended, not replaced. Priority beats still sort first, requires edges still take precedence. Only the tie-breaking within equal-priority, equal-constraint beats changes.
- Beats without `dilemma_impacts` fall back to the previous alphabetical behavior.

### DIFF
--- a/src/questfoundry/graph/grow_algorithms.py
+++ b/src/questfoundry/graph/grow_algorithms.py
@@ -219,13 +219,18 @@ def topological_sort_beats(
 ) -> list[str]:
     """Topologically sort a subset of beats using requires edges.
 
-    Uses Kahn's algorithm with deterministic tie-breaking. When multiple beats
-    have in-degree 0, beats in *priority_beats* sort first, then alphabetically.
-    This lets callers push shared/common beats early in the sequence and
-    exclusive beats late, so arcs naturally diverge toward the end.
+    Uses Kahn's algorithm with deterministic tie-breaking.  Tie-breaking
+    order (highest priority first):
+
+    1. **Priority tier** — beats in *priority_beats* (shared) before others.
+    2. **Dilemma round-robin** — beats from the dilemma that has emitted
+       fewest beats so far sort first, interleaving dilemmas instead of
+       clustering all of one dilemma's beats together.
+    3. **Alphabetical** — deterministic within the same tier and count.
 
     Args:
-        graph: Graph containing requires edges.
+        graph: Graph containing requires edges and beat nodes with
+            optional ``dilemma_impacts`` metadata.
         beat_ids: Subset of beat node IDs to sort.
         priority_beats: Optional set of beat IDs that should sort before
             non-priority beats when topological constraints allow. When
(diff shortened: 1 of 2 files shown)


pvliesdonk:
Addressed the test DRY suggestion — extracted `_create_dilemma_chain` helper in 6008083.

On the performance concern: the queue re-sort on every iteration was already present in the original code (before this PR). My change only adds a dict lookup to the sort key. In practice, V is 10–30 beats — O(V^2 log V) at that scale is sub-millisecond. A meta-queue approach would be over-engineering for the actual data sizes. If beat counts ever grow significantly, we can revisit.

PR #937, opened 2026-02-17
fix(tracing): prevent SQLite threading errors in langsmith background serialization

pvliesdonk:
## Problem

SQLite threading errors during pipeline execution:
```
ProgrammingError: SQLite objects created in a thread can only be used
in that same thread. The object was created in thread id X and this is thread id Y.
```

Persists even after upgrading langsmith to 0.7.3 (#935).

## Root Cause

Langsmith's `@traceable` decorator serializes function **inputs** (including `self` and `model`) in a background thread. Stage methods like `GrowStage.execute(self, model, ...)` pass objects that hold or can reach SQLite connections. When langsmith's background serialization thread walks these objects, it triggers Python's SQLite thread-safety check.

Langsmith itself doesn't use SQLite — the error comes from our objects being serialized in langsmith's thread.

## Changes

Add `process_inputs` callback to our `traceable` wrapper in `src/questfoundry/observability/tracing.py`:
- Primitive types (str, int, float, bool, None) pass through unchanged
- Collections get string-truncated representations
- All other objects (stage instances, LangChain models, callbacks) get replaced with `<TypeName>` to avoid deep serialization

The trace still records *what* was passed without touching thread-unsafe resources.

## Test Plan

- `uv run mypy` and `uv run ruff check` pass
- Manual verification: run a pipeline with `LANGSMITH_TRACING=true` and confirm no SQLite threading errors

## Risk / Rollback

Low risk — only affects trace input serialization, not pipeline behavior. Traces will show type names instead of full object serializations for non-primitive inputs.

### DIFF
--- a/src/questfoundry/observability/tracing.py
+++ b/src/questfoundry/observability/tracing.py
@@ -130,7 +130,7 @@ def _safe_serialize_inputs(inputs: dict[str, Any]) -> dict[str, Any]:
     handlers) trigger ``ProgrammingError: SQLite objects created in a
     thread can only be used in that same thread``.
 
-    This function replaces such objects with their ``repr()`` so the
+    This function replaces such objects with a ``<TypeName>`` placeholder so the
     trace still records *what* was passed without touching thread-unsafe
     resources.
     """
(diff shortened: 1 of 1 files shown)


pvliesdonk:
Addressed all review comments in ce527b7:

- **list/tuple bug** (Claude, Gemini): replaced `str(value)[:500]` with `f"<{type(value).__name__}[{len(value)}]>"` — no `str()` call on elements
- **dict values bug** (Claude, Gemini): dict values now use primitive pass-through for safe types, `<TypeName>` for everything else
- **docstring** (Claude, Gemini): updated to say `<TypeName>` placeholder instead of `repr()`
- **inline comment** (Gemini): updated to match implementation

Issue #914, opened 2026-02-15
feat(seed/grow): add residue_weight primitive for mid-story prose control


pvliesdonk:
## Parent Epic
Part of #911 (Topology/Prose Layer Separation) — Phase 3

## Problem

Currently, mid-story prose variation is controlled indirectly by `convergence_policy`:
- `hard` → no shared beats (too aggressive)
- `soft` → residue beats after payoff budget (no control over prose impact)
- `flavor` → overlays only (too weak — cosmetic state changes can't carry narrative weight)

There's no way to say "this choice should affect prose at convergence points but not endings" or "this choice is cosmetic — overlays only, shared prose must ignore it."

## Solution

Add `residue_weight` to `DilemmaAnalysis` as the prose-layer control for mid-story passages:

| Value | Meaning | Obligation |
|-------|---------|-----------|
| `heavy` | **Requires** explicit variant (residue beat or passage variant) at/near convergence | FILL must write state-specific prose |
| `light` | Shared prose allowed if state-neutral; use variant only when neutrality is hard | GROW may create variants; FILL tries neutral prose first |
| `cosmetic` | Overlays/metadata only; **shared prose MUST ignore** this choice | Negative obligation — FILL must not reference it |

### Relationship to existing fields

```
convergence_policy  → topology (beat sharing)
ending_salience     → prose layer for endings (Phase 1)
residue_weight      → prose layer for mid-story (this phase)
```

### 1. Model change (~20 lines)
**File:** `src/questfoundry/models/seed.py`

```python
# Add to DilemmaAnalysis
residue_weight: Literal["heavy", "light", "cosmetic"] = "light"
```

Default `"light"` — most dilemmas have some mid-story impact but don't require dedicated variants.

### 2. SEED prompt change (~25 lines)
**File:** `prompts/templates/seed.md` (Section 7)

Add `residue_weight` field with guidance:
- "`heavy`: reader must feel this choice in shared passages — creates explicit variants"
- "`light`: default — shared prose should be state-neutral where possible"
- "`cosmetic`: purely visual/tonal — handled by overlays, prose ignores it"

### 3. GROW residue candidate filtering (~30 lines)
**File:** `src/questfoundry/graph/grow_algorithms.py`

Update `find_residue_candidates()` and `create_residue_passages()`:
- `heavy` → always create residue variant at convergence point
- `light` → create variant only if LLM flags the passage as hard to write neutrally (existing behavior)
[...]

PR #939, opened 2026-02-17
fix(validation): filter prose_neutrality to actually-diverging dilemmas

### LINKED ISSUE #938: fix(validation): prose_neutrality check doesn't filter for actually-diverging dilemmas
## Bug

`check_prose_neutrality` (added in #921, closing #915) flags shared passages for **all** dilemmas present in their covering arcs, rather than only dilemmas where those arcs **actually diverge** (i.e., chose different paths for that dilemma).

## Reproduction

Run `qf grow` on a project with combinatorial arc structure (multiple dilemmas, each with 2 paths). Every shared passage triggers failures for every `heavy/high` dilemma, even when all covering arcs chose the same path for that dilemma at that point.

Observed in `projects/test-new`: 116 failures + 238 warnings, all `prose_neutrality`. Every structural check passes.

## Root Cause

`grow_validation.py:1433-1437` — the comment says "find dilemmas that diverge across covering arcs" but the code computes a **union** of all dilemma IDs without checking whether arcs actually differ on their path choice for each dilemma:

```python
# Comment: "dilemma appears in some arcs but not all = paths diverged before this beat"
# Actual code: union of ALL dilemmas from ALL covering arcs
all_dilemma_ids: set[str] = set()
for arc_id in covering_arcs:
    all_dilemma_ids.update(arc_dilemmas.get(arc_id, set()))
```

When every arc contains all 7 dilemmas (combinatorial structure), this means every shared passage is checked against every dilemma — producing `58 passages x 2 heavy/high dilemmas = 116 false failures`.

## Expected Behavior

Only flag a passage for a dilemma when the covering arcs **chose different paths** for that dilemma. If all covering arcs agree on the same path for dilemma X, there is no prose-neutrality concern for X at that passage.

## Fix

After collecting `all_dilemma_ids`, filter to only dilemmas where the covering arcs have 2+ distinct path choices for that dilemma. This matches the intent described in #915: "shared passage reachable from arcs **differing** on a heavy residue or high ending salience dilemma".

## Contributing Factor

PR #934 made `ending_salience`/`residue_weight` required fields in the seed stage. Before that, small models that omitted them got safe `light/low` defaults, masking this bug. Now that models explicitly set `heavy/high` where appropriate, the false positives surface.

Parent epic: #911

pvliesdonk:
## Problem

`check_prose_neutrality` (added in #921) has two issues that cause the GROW stage to fail on valid projects:

1. **False positives from non-diverging dilemmas**: Flags shared passages for *all* dilemmas present in covering arcs, rather than only dilemmas where arcs **actually diverge** (chose different paths). In combinatorial arc structures this produces massive false positives — 116 spurious failures in `projects/test-new`.

2. **Severity too strict for current grow capabilities**: Even after filtering to genuinely-diverging dilemmas, the remaining `heavy/high` checks produce `fail` severity — but the grow stage can't yet wire mid-story variant routing (`split_and_reroute` only handles endings). This blocks the pipeline on valid dilemma configurations.

Fixes #938

### Commit 1: Filter to actually-diverging dilemmas
- **`grow_validation.py`**: After collecting `all_dilemma_ids` from covering arcs, added a `diverging_dilemmas` filter that only retains dilemmas where covering arcs have 2+ distinct path choices
- **`test_grow_validation.py`**: Added `test_non_diverging_dilemma_not_flagged` and `test_only_diverging_dilemma_flagged_in_multi_dilemma` regression tests

### Commit 2: Downgrade heavy/high severity to warn
- **`grow_validation.py`**: Changed `severity="fail"` to `severity="warn"` for heavy/high without routing, with comment explaining the grow stage limitation
- **`test_grow_validation.py`**: Updated test names and assertions from `fails` → `warns`

## Not Included / Future PRs

- Mid-story variant routing for heavy/high dilemmas: tracked in #914
- Once #914 is implemented, severity should be upgraded back to `fail`

## Test Plan

```
uv run pytest tests/unit/test_grow_validation.py::TestCheckProseNeutrality -x -q
# 10 passed (8 existing + 2 new)
uv run mypy src/questfoundry/graph/grow_validation.py  # clean
uv run ruff check src/ tests/                          # clean
```

Verified against `projects/test-new/graph.db`:
- Before: 116 failures, 238 warnings → pipeline blocked
- After commit 1: 48 failures, 205 warnings (eliminated 68 false positives)
- After commit 2: **0 failures**, 253 warnings → pipeline unblocked

## Risk / Rollback

Low risk — validation-only changes. No impact on graph mutations or stage execution. The divergence filter strictly reduces false positives. The severity downgrade is a conscious trade-off documented in #914.

### DIFF
--- a/src/questfoundry/graph/grow_validation.py
+++ b/src/questfoundry/graph/grow_validation.py
@@ -1462,10 +1462,15 @@ def check_prose_neutrality(graph: Graph) -> list[ValidationCheck]:
                 continue  # Variant routing exists, prose contract met
 
             if weight == "heavy" or salience == "high":
+                # Warn rather than fail: the grow stage does not yet
+                # wire mid-story variant routing for heavy/high
+                # dilemmas (split_and_reroute handles endings only).
+                # Tracked as a future capability; blocking the pipeline
+                # on an unimplemented feature produces false failures.
                 checks.append(
                     ValidationCheck(
                         name="prose_neutrality",
-                        severity="fail",
+                        severity="warn",
                         message=(
                             f"Shared passage {pid} requires variant routing "
                             f"for dilemma '{label}' "
--- a/src/questfoundry/graph/grow_validation.py
+++ b/src/questfoundry/graph/grow_validation.py
@@ -1430,15 +1430,29 @@ def check_prose_neutrality(graph: Graph) -> list[ValidationCheck]:
         if len(covering_arcs) < 2:
             continue  # Not shared
 
-        # Find dilemmas that diverge across covering arcs
-        # (dilemma appears in some arcs but not all = paths diverged before this beat)
+        # Find dilemmas that actually diverge across covering arcs:
+        # a dilemma diverges only when covering arcs chose *different* paths
+        # for it (i.e., 2+ distinct path IDs for that dilemma).
         all_dilemma_ids: set[str] = set()
         for arc_id in covering_arcs:
             all_dilemma_ids.update(arc_dilemmas.get(arc_id, set()))
 
+        diverging_dilemmas: set[str] = set()
+        for dilemma_id in all_dilemma_ids:
+            paths_for_dilemma: set[str] = set()
+            for arc_id in covering_arcs:
+                for path_raw in arc_nodes.get(arc_id, {}).get("paths", []):
+                    p_id = normalize_scoped_id(path_raw, "path")
+                    p_data = path_nodes.get(p_id, {})
+                    raw_did = p_data.get("dilemma_id", "")
+                    if raw_did and normalize_scoped_id(raw_did, "dilemma") == dilemma_id:
+                        paths_for_dilemma.add(p_id)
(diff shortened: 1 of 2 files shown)


pvliesdonk:
Addressed review comments:

- Optimized diverging_dilemmas logic to O(A*p) with a dilemma->paths map and documented the one-path-per-dilemma-per-arc invariant (commit 0e74120).
- Removed unused path::p2 from the fixture (commit 0e74120).
- PR description now explicitly mentions the heavy/high severity downgrade and references #914.
- On the spine-contains-all-paths comment: arcs are expected to contain at most one path per dilemma (enumerate_arcs invariant). Added a comment documenting this assumption; no malformed-data test added unless we decide to allow multi-path arcs.

pvliesdonk:
Added a small regression test for missing dilemma_id paths to improve diff coverage (commit 7df0f2a). Codecov/patch should now pass once CI reruns.

Issue #970, opened 2026-02-22
fix(grow): apply_intersection_mark() violates hard convergence policy


pvliesdonk:
## Problem

GROW stage validation fails with 13 `convergence_policy_compliance` errors for the `tide_walkers_sacrifice_or_silent_accord_expose` dilemma (which has `convergence_policy=hard`). All failures are: "hard policy violated — 2 shared beat(s) after divergence".

This persisted through Epic #911 (Topology/Prose separation) and Epic #950 (Unified Routing Plan) because both addressed downstream symptoms rather than the upstream graph construction bug.

## Root Cause (from Discussion #965)

Multi-agent deliberation identified the PRIMARY BUG:

**Location:** `apply_intersection_mark()` in `src/questfoundry/graph/grow_algorithms.py` (lines 3306-3351)

**Issue:** Cross-assigns `belongs_to` edges from ALL beats to ALL paths in an intersection WITHOUT checking `convergence_policy`.

**How it breaks:**
1. Beat A from soft path 1 + Beat B from hard path 2 are in same intersection
2. Function collects ALL path IDs: `{soft_path_1, hard_path_2}`
3. Adds `belongs_to` edges: Beat A → hard_path_2, Beat B → soft_path_1
4. Result: Hard dilemma beats become shared across mutually exclusive choices ❌

**Evidence from test-new:**
- Beat `compass_beneficial_or_cursed__beneficial_beat_03` has `belongs_to` edges to BOTH a soft path AND the hard path `tide_walkers...silent_accord_expose`
- Arcs with opposite hard choices share these cross-assigned beats
- Validation correctly detects this as violating the hard convergence contract

## Fix

Add convergence policy check to `apply_intersection_mark()`:

```python
# Before adding cross-path belongs_to edges:
# 1. Group paths by dilemma
# 2. For each dilemma group:
#    - Query convergence_policy for the dilemma
#    - If convergence_policy == "hard": Do NOT cross-assign within group
#    - If convergence_policy == "soft": Allow cross-assignment (current behavior)
# 3. Cross-assign only between dilemma groups (safe - different dilemmas can share)
```

**Expected outcome:**
- Hard dilemma paths remain isolated (no shared beats after divergence)
- Soft dilemma paths can still share beats (efficiency preserved)
- Arc enumeration UNION will naturally exclude hard-path beats from wrong arcs
- All 13 convergence_policy_compliance failures should resolve

## Implementation Scope

[...]

Issue #950, opened 2026-02-19
Epic: Unified Routing Plan (ADR-017)


pvliesdonk:
## Overview

Implement the Unified Routing Plan architecture decided in [ADR-017](https://github.com/pvliesdonk/questfoundry/pull/949) and [Discussion #948](https://github.com/pvliesdonk/questfoundry/discussions/948).

Replace the current incremental 3-phase graph mutation (Phase 15 → 21 → 23) with a plan-then-execute architecture that computes all routing needs globally before applying any mutations.

## Background

Since implementing the Topology/Prose Layer Separation (Epic #911), GROW has never produced a fully successful run. Multi-agent deliberation in Discussion #948 identified 8 bugs and reached unanimous consensus on the fix architecture.

## Track 1 — Tactical Fixes (unblock GROW, ~60 lines)

These fixes are independent and can be implemented in parallel:

- [ ] **T1: Fix `routed_passages` semantic inversion** (~20 lines) — Build from `residue_for`/`family_codewords` metadata instead of `from_passage`
- [ ] **T2: Fix `build_arc_codewords` codeword scope** (~15 lines) — Add `scope` parameter
- [ ] **T3: Switch to `keep_fallback=True` + fallback-aware CE** (~11 lines) — Never delete original edges
- [ ] **T4: Fix per-dilemma guard in `find_heavy_divergence_targets`** (~15 lines) — Per-(passage, dilemma) instead of per-passage
- [ ] **T5: Move Phase 15 priority after choices exist** (~2 lines) — Priority 15 → ~20.5

## Track 2 — Strategic Refactor (prevent recurrence, ~800 lines)

Sequential dependency chain (each depends on prior):

- [ ] **S1: `RoutingPlan` dataclass + `compute_routing_plan()`** (~300 lines) — Pure function, no side effects
- [ ] **S2: Wire Phase 15 to produce plan entries** (~200 lines) — Advisory LLM, no direct mutations
- [ ] **S3: Collapse Phases 21+23 into `apply_routing_plan()`** (~200 lines) — Single atomic application
- [ ] **S4: Validation alignment** (~100 lines) — Plan-aware CE/ME checks

## Dependencies

- Track 2 depends on Track 1 being merged
- S1 → S2 → S3 → S4 (sequential)
- Supersedes remaining work from Epic #911 (Phase 3 #914)

## References

- ADR-017: PR #949
- Deliberation: Discussion #948
- Original diagnosis: Discussion #910
[...]

Issue #911, opened 2026-02-15
epic: Topology/Prose Layer Separation


pvliesdonk:
## Problem

The current convergence system conflates graph topology (beat sharing) with prose variation (how choices affect narrative). This causes:

- **1280 ending families** from 8 binary dilemmas (#909) — every codeword contributes to signatures even when narratively irrelevant
- **Ambiguity trap** — shared passages that must simultaneously serve multiple story states, degrading prose quality
- **No mid-story prose control** — `convergence_policy` only governs topology; there's no way to say "this choice affects prose but not structure"
- **Two incompatible variant patterns** — `split_ending_families()` uses a hub pattern (extra hops), `_redirect_edges_for_merge()` uses incoming-edge rewrites

## Solution: Topology/Prose Layer Separation

Separate structural convergence (topology) from prose variation decisions (prose layer):

| Layer | Controls | Fields | Used by |
|-------|----------|--------|---------|
| **Topology** | Beat sharing, path reconvergence | `convergence_policy` (existing) | GROW phases 6-9 |
| **Prose** | Which choices affect written prose | `ending_salience`, `residue_weight` (new) | GROW ending splits, residue beats, FILL context |

### Key Design Decisions (from Discussion #910)

1. **`ending_salience: high | low | none`** on DilemmaAnalysis — controls which dilemmas drive ending differentiation
2. **`residue_weight: heavy | light | cosmetic`** on DilemmaAnalysis — controls mid-story prose variation
3. **Negative obligations** — `low/none` and `cosmetic` are explicit contracts: "shared prose MUST NOT depend on this choice"
4. **Unified variant routing primitive** — incoming-edge rewriting (not hub pattern) for all variant types
5. **CE+ME validation** — collectively-exhaustive + mutually-exclusive routing validation for all variant choice sets

### Phased Implementation

```
Phase 1 (#912) ← no deps
    ↓
Phase 2 (#913) ← depends on #912
    ↓
Phase 3 (#914) ← depends on #912, #913
    ↓
Phase 4 (#915) ← depends on #912, #913, #914
    ↓
Phase 5 (#916) ← depends on #912-#915
```

- [ ] **Phase 1:** `ending_salience` primitive — #912 (closes #909) ~200-300 lines
- [ ] **Phase 2:** Unified variant routing — #913 ~400-500 lines
- [ ] **Phase 3:** `residue_weight` primitive — #914 ~250-350 lines
[...]

PR #971, opened 2026-02-22
fix(grow): respect convergence_policy in apply_intersection_mark()

pvliesdonk:
## Summary

Fixes the root cause of GROW stage validation failures identified in Discussion #965 multi-agent deliberation.

## Problem

GROW stage fails with 13 `convergence_policy_compliance` errors for hard dilemmas. The `apply_intersection_mark()` function was cross-assigning `belongs_to` edges from ALL beats to ALL paths in an intersection WITHOUT checking `convergence_policy`.

**Result:** Beats from hard dilemma paths became shared across mutually exclusive choices, violating the hard convergence contract.

## Changes

Modified `apply_intersection_mark()` in `src/questfoundry/graph/grow_algorithms.py` to:

1. Group paths by dilemma (via 'offers' edges)
2. Query each dilemma's `convergence_policy`
3. **Deny** cross-path `belongs_to` assignment within same hard dilemma
4. **Allow** cross-assignment for soft/flavor dilemmas (preserves efficiency)
5. **Allow** cross-assignment between different dilemmas (always safe)

## Test Plan

**Manual testing on projects/test-new:**
- Before: 13 convergence_policy_compliance errors
- After: Expect 0 errors (all 13 should resolve)

**Command:**
```bash
uv run qf grow --project projects/test-new
```

**Expected outcome:**
- Hard dilemma paths remain isolated (no shared beats after divergence)
- Soft dilemma paths can still share beats (efficiency preserved)
- Validation passes (0 convergence_policy_compliance failures)

## Risk / Rollback

**Low risk:**
- Surgical fix to single function (20 lines added)
- Only affects intersection construction behavior
- Preserves existing soft/flavor dilemma sharing

**Rollback:** Revert this commit if validation still fails or new issues appear.

## Context

- **Closes:** #970
- **Discussion:** #965 (multi-agent deliberation with Claude Opus 4.6, Gemini 3 Pro, GPT-5.2)
- **Related:** Epic #911 (Topology/Prose separation), Epic #950 (Unified Routing Plan)
- **Design rationale:** Hard dilemmas prevent 2^n ending explosion (Discussion #910, #948)

### DIFF
--- a/src/questfoundry/graph/grow_algorithms.py
+++ b/src/questfoundry/graph/grow_algorithms.py
@@ -3364,18 +3364,29 @@ def apply_intersection_mark(
             bid_dilemmas = {path_to_dilemma.get(p) for p in bid_paths if p in path_to_dilemma}
             target_dilemma = path_to_dilemma.get(path_id)
 
-            # Allow cross-assignment if:
-            # 1. Different dilemmas (always safe)
-            # 2. Same dilemma with soft/flavor policy (allows sharing)
-            # Deny if same dilemma with hard policy (requires isolation)
-            if target_dilemma and target_dilemma in bid_dilemmas:
-                # Same dilemma - check policy
-                policy = dilemma_to_policy.get(target_dilemma, "soft")
-                if policy == "hard":
-                    # Hard policy - do NOT cross-assign within dilemma
-                    continue
+            # Deny cross-assignment if target path is from a hard dilemma
+            # Hard dilemmas must have exclusive beats (no sharing with ANY other path)
+            target_policy = (
+                dilemma_to_policy.get(target_dilemma, "soft") if target_dilemma else "soft"
+            )
+            if target_policy == "hard":
+                # Hard policy - do NOT cross-assign to this path
+                continue
 
-            # Safe to cross-assign
+            # Also deny if current beat belongs to any hard dilemma path
+            beat_from_hard = False
+            for bid_dilemma in bid_dilemmas:
+                if bid_dilemma:
+                    bid_policy = dilemma_to_policy.get(bid_dilemma, "soft")
+                    if bid_policy == "hard":
+                        # Beat from hard dilemma - do NOT cross-assign
+                        beat_from_hard = True
+                        break
+
+            if beat_from_hard:
+                continue
+
+            # Safe to cross-assign (both source and target are soft/flavor)
             new_edges.append((bid, path_id))
 
     # Update each beat's node data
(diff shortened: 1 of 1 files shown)


pvliesdonk:
## Update: Initial Fix Incomplete

After testing on `projects/test-new`, the fix is **still failing** with the same 12 convergence_policy_compliance errors.

### Root Cause Analysis (Deeper Investigation)

The problem is **NOT just in `apply_intersection_mark()`** (though that's part of it).

**The actual issue is in arc enumeration (lines 1391-1393):**

```python
beat_set: set[str] = set()
for pid in path_combo:
    beat_set.update(path_beat_sets.get(pid, set()))  # UNION without filtering
```

This creates arcs by taking the UNION of all beats from all paths, **without considering convergence_policy**.

**What happens:**
1. Spine arc = all canonical paths (one per dilemma)
2. Hard dilemma canonical path contains beats (including gap beats)
3. Those beats end up in spine sequence via UNION
4. Branch arc chooses NON-canonical path of hard dilemma
5. Branch arc shares some beats with spine (gap beats, beats before divergence)
6. Validation correctly rejects this as hard policy violation

**My fix to `apply_intersection_mark()` doesn't address this** because arc sequences are built from UNION logic, not from `belongs_to` edges.

### The Nuanced Requirement (from User)

> "Beat-sharing can only happen for non-hard and/or before commit"

This means:
- Beats BEFORE commit point of hard dilemmas → sharing OK
- Beats AFTER commit point of hard dilemmas → sharing FORBIDDEN
- Beats from soft/flavor dilemmas → sharing always OK

**This is temporal, not categorical.**

### Next Steps

The fix needs to be in **arc enumeration logic** (lines 1391-1393), not just in intersection marking:

1. When building arc beat_set via UNION
2. Check if any path in this combo has a hard dilemma
3. For hard dilemmas, only include beats BEFORE the commit point
4. For soft dilemmas, include all beats (current behavior)

This is more complex than the current fix. Requesting guidance before proceeding.

### Questions

1. What exactly is the "commit point" in the graph structure?
2. Is it the `diverges_at` field on arcs?
3. Or is it per-dilemma (first beat after paths diverge)?
4. Should I revert my `apply_intersection_mark()` changes and focus on arc enumeration instead?

pvliesdonk:
Closing as superseded. The approach in this PR (patching `apply_intersection_mark()` to check convergence policy) addresses a symptom, not the root cause.

Documents 1 and 3 (merged in PRs #973 and #976) redefine the intersection model entirely — intersections become co-occurrence groups rather than cross-path `belongs_to` edges, and `convergence_policy` is replaced by `dilemma_role`. The fix attempted here will be subsumed by the intersection model migration planned in Discussion #980.

See Issue #977 for the migration tracking.

PR #1099, opened 2026-03-02
fix: remove consequence_ids from path nodes, use has_consequence edges

### LINKED ISSUE #1097: PARTIAL: consequence_ids field on path nodes always empty despite correct has_consequence edges
## Problem

Path nodes have a `consequence_ids` field that is **always `[]`** at runtime, despite `has_consequence` edges being correctly created (14 confirmed in `projects/test-new/graph.db`). The field is redundant with the edges but misleading when empty — any code reading `path.consequence_ids` sees an empty list and may incorrectly infer no consequences exist.

## Decision Required

Two options:

**Option A — Populate the field**: After creating `has_consequence` edges, backfill `consequence_ids` on the path node to match. This adds redundancy but satisfies any consumers expecting the field.

**Option B — Remove the field**: Remove `consequence_ids` from the path node model and update all consumers to use `has_consequence` edges instead. This eliminates the inconsistency at the cost of a graph traversal.

Option B is preferred: removing the redundant field is cleaner and avoids keeping two sources of truth in sync.

## Changes Required (if Option B)

1. Remove `consequence_ids` from the path node Pydantic model in `seed.py`
2. Search for and update all consumers that read `consequence_ids` from path nodes
3. Verify no path nodes have `consequence_ids` in the database after the change

## Changes Required (if Option A)

1. After inserting `has_consequence` edges, update the path node's `consequence_ids` field to match
2. Add a test asserting that `consequence_ids` is non-empty when `has_consequence` edges exist

## Acceptance Criteria

- [ ] No path nodes have `consequence_ids: []` when `has_consequence` edges exist for that path
- [ ] Either the field is removed and consumers use edges, OR the field is populated to match edges
- [ ] No two sources of truth exist for consequence membership on paths

## Verification (if Option B — removal)

```bash
# Should return 0 (field removed from all path nodes)
sqlite3 projects/<project>/graph.db \
  "SELECT count(*) FROM nodes WHERE type='path' AND data LIKE '%consequence_ids%'"

# Should return 0 (no code reading the removed field)
grep -rn "consequence_ids" src/questfoundry/
```

## Verification (if Option A — populate)

```bash
# Should return 0 (no path nodes with empty consequence_ids when edges exist)
sqlite3 projects/<project>/graph.db \
  "SELECT count(*) FROM nodes WHERE type='path' AND data LIKE '%\"consequence_ids\": []%'"
```

pvliesdonk:
## Problem

The `Path` model and path node data stored a redundant `consequence_ids` field alongside the `has_consequence` edges in the graph. This created a dual source of truth: consequences were reachable both via graph traversal (`has_consequence` edges) and via the stored field on the path node. The `has_consequence` edges are the canonical representation per the ontology.

## Changes

- Removed `consequence_ids` field from the `Path` model in `src/questfoundry/models/seed.py`
- Removed `consequence_ids` from path node data written in `src/questfoundry/graph/mutations.py`
- Updated `src/questfoundry/graph/context.py` to look up consequences by matching `path_id` from `seed_output.consequences` directly instead of reading `consequence_ids` from the stored node
- Removed corresponding test assertions that verified the now-deleted field

## Not Included / Future PRs

None — this is a self-contained removal.

## Verification

```bash
grep -rn "consequence_ids" src/questfoundry/
```

Returns 0 matches — the field is fully removed from the codebase.

## Test Plan

- `uv run pytest tests/unit/test_mutations.py tests/unit/test_graph_context.py tests/unit/test_serialize.py -x -q`
- All tests pass with the updated assertions

## Risk / Rollback

Low risk — removes a redundant stored field. All consequence lookups now go through `seed_output.consequences` matching on `path_id`, which is the canonical source. No external consumers of `consequence_ids` exist in this codebase.

Closes #1097

### DIFF
--- a/src/questfoundry/graph/context.py
+++ b/src/questfoundry/graph/context.py
@@ -954,10 +954,11 @@ def format_dilemma_analysis_context(
         d_raw = strip_scope_prefix(p.dilemma_id)
         paths_by_dilemma.setdefault(d_raw, []).append(p)
 
-    # Build consequence lookup: normalized raw_id → Consequence
-    cons_by_id: dict[str, Consequence] = {}
+    # Build consequence lookup: normalized path raw_id → list of Consequences
+    cons_by_path: dict[str, list[Consequence]] = {}
     for c in seed_output.consequences:
-        cons_by_id[strip_scope_prefix(c.consequence_id)] = c
+        p_raw = strip_scope_prefix(c.path_id)
+        cons_by_path.setdefault(p_raw, []).append(c)
 
     dilemma_blocks: list[str] = []
     for d in sorted(seed_output.dilemmas, key=lambda x: x.dilemma_id):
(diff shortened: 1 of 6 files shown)


pvliesdonk:
## Design Conformance

**Skip exception applies** — pure internal refactoring, no behavioral change to LLM-facing output.

`consequence_ids` was a redundant stored field duplicating information already in `has_consequence` edges. The context builder now reads those edges directly instead, producing identical consequences in LLM context. Per CLAUDE.md skip exception: pure refactoring with no behavioral change.

PR #1102, opened 2026-03-02
fix: strengthen location_alternatives prompt and enrich context with cross-path locations

### LINKED ISSUE #1092: DEAD: Entity flexibility / location_alternatives never produced by LLM — GROW intersection discovery broken
## Problem

The SEED stage has full infrastructure for entity flexibility and `location_alternatives` on beats — schema (`seed.py:298-301`), mutation code (`mutations.py:1834-1839, 1877-1879`), and prompt text (`serialize_seed_sections.yaml:436, 451`) all exist. However, runtime evidence from `projects/test-new/` shows:

- **0 `flexibility` edges** in `graph.db`
- **0 beats** with `location_alternatives` populated

This breaks GROW's intersection discovery: the spec states that beats sharing location alternatives are "the raw material for intersection." Without this data, GROW cannot find valid intersection candidates via location overlap.

Design doc references:
- `docs/design/how-branching-stories-work.md` Part 3 ("the raw material for intersection")
- `docs/design/document-3-ontology.md` Part 3

## Root Cause (hypothesis)

The serialize prompt likely fails to explain what `location_alternatives` is for or provide examples that help the LLM understand when to populate it. The `prompt-engineer` subagent must audit the prompt before any fix is merged.

## Changes Required

1. Audit `prompts/templates/serialize_seed_sections.yaml` (around lines 436, 451) for the `location_alternatives` instruction
2. Use the `prompt-engineer` subagent to redesign the instruction with concrete examples (e.g., "the market district could also be the harbor")
3. Run SEED and inspect `logs/llm_calls.jsonl` to confirm the LLM is producing the field
4. Verify `flexibility` edges appear in `graph.db`

## Acceptance Criteria

- [ ] After running SEED, at least some beats have `location_alternatives` populated in `graph.db`
- [ ] `flexibility` edges appear in `graph.db` (count > 0)
- [ ] `prompt-engineer` subagent has reviewed the serialize prompt changes before merge
- [ ] Design conformance: `architect-reviewer` sign-off with 0 DEAD findings for entity flexibility

## Verification

```bash
# Should return > 0 after a SEED run
sqlite3 projects/<project>/graph.db \
  "SELECT count(*) FROM edges WHERE type='flexibility'"

# Should return > 0
sqlite3 projects/<project>/graph.db \
  "SELECT count(*) FROM nodes WHERE data LIKE '%location_alternatives%'"
```

## Not Included / Future PRs

- Expanding flexibility beyond locations to characters/factions/objects (tracked separately; see related PARTIAL issue)
- GROW changes to consume flexibility edges for intersection discovery (follow-on after data is reliably produced)

pvliesdonk:
## Problem
Issue #1092: `location_alternatives` is DEAD — the field always defaults to `[]` because there is no explanation of what GROW uses it for, and the empty-array default signals "skip" to the model.

## Changes
- Added explanation in prompt that GROW uses `location_alternatives` for intersection detection between paths — giving the LLM a reason to populate the field
- Added a concrete named example with multiple plausible locations
- Changed framing to "list locations where the beat could plausibly move without breaking scene logic"
- `serialize.py` now injects sibling path locations as context so the LLM knows what locations other paths use, enabling informed choices that maximise intersection opportunities
- Files: `prompts/templates/serialize_seed_sections.yaml`, `src/questfoundry/agents/serialize.py`

## Not Included / Future PRs
- GROW intersection detection logic is unchanged (already consumes the field correctly)

## Test Plan
- Prompt + context injection change
- Verify via `logs/llm_calls.jsonl`: inspect beat objects for non-empty `location_alternatives` lists, and confirm sibling location context appears in the `messages` array

## Risk / Rollback
- Low risk: no schema change; existing `[]` values remain valid
- Context injection adds a small token overhead per SEED call

Closes #1092

### DIFF
--- a/prompts/templates/serialize_seed_sections.yaml
+++ b/prompts/templates/serialize_seed_sections.yaml
@@ -445,10 +445,41 @@ beats_prompt: |
 
   **Use IDs from YOUR story's manifest** - the examples above show the FORMAT only.
 
+  ## location_alternatives — GROW intersection hint (IMPORTANT)
+
+  GROW uses `location_alternatives` to find beats from DIFFERENT paths that could
+  share a scene at the same physical location. If your beat at `location::harbor`
+  lists `location_alternatives: ["location::market_district"]`, GROW knows this
+  beat could be relocated to the market district and thus co-occur with a path-B
+  beat that is already set there.
+
+  Think of it as: "Where ELSE could this scene plausibly take place without
+  breaking its logic?" List every location from the Entity IDs that would make
+  narrative sense as an alternative setting. Leave it empty ONLY when the beat
+  is so specific to its primary location that no other setting could work.
+
+  EXAMPLE (format — use your story's actual location IDs):
+  A beat where two characters argue about a stolen document is set at
+  `location::archive_hall`. It could equally happen at `location::council_chamber`
+  or `location::private_study` — so list those as alternatives. A beat where a
+  character jumps from a specific cliff has no viable alternative location, so
+  leave `location_alternatives` empty.
+
+  ```json
+  "location": "location::archive_hall",
+  "location_alternatives": ["location::council_chamber", "location::private_study"]
+  ```
+
+  The brief includes a "### Sibling path locations" section listing which
+  locations OTHER paths use. Prefer listing those as alternatives when they
+  fit — that is exactly where GROW will look for intersection opportunities.
+
   ## Rules
   - effect must be exactly "advances", "reveals", "commits", or "complicates"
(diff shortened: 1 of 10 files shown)


pvliesdonk:
**Deferred item (finding #4 — hidden prompt):** The sibling path section is built inline alongside dynamically generated per-sibling lines (path ID, name, description). Extracting only the static intro to YAML while leaving the per-sibling loop in Python would split the block without eliminating the concern. Filed for future consideration when we refactor the beat context builders.

PR #1126, opened 2026-03-03
fix: SEED hard-fails on linear story and missing post-commit beats

### LINKED ISSUE #1122: fix(seed): hard-fail on linear story and missing post-commit beats
### Problem

Two warnings during SEED commit are currently silent degradations that allow a broken story structure to proceed into GROW:

1. **`Only 1 arc(s) — this is a linear story`** (`arc_count < 2`): A story with fewer than 2 arcs is not interactive fiction — it has no meaningful player choices. GROW then runs on a structure that cannot produce a branching narrative.

2. **`seed_scaffold_warning` — path has no beat after commit beat**: An arc without consequence beats is structurally incomplete. The post-commit beat is where the player sees the result of their choice. Its absence means the arc ends at the decision point with no payoff.

Both are detected at SEED commit time in `mutations.py` but logged as warnings and allowed to proceed.

### Root Cause

Two separate failures:
- **Detection**: The checks exist but use `log.warning` instead of raising errors.
- **Prevention**: The SEED serialize prompt (`prompts/templates/serialize_seed_sections.yaml`) does not require a minimum number of player-facing dilemmas, nor explicitly require a post-commit consequence beat on every path.

### Fix

**Part 1 — Promote to hard errors** (`mutations.py`):
- `arc_count < 2` → raise `SeedValidationError` (or equivalent), block SEED from committing
- `seed_scaffold_warning` (no post-commit beat) → same, raise and block

**Part 2 — Prevent via prompt** (`prompts/templates/serialize_seed_sections.yaml`):
- Add explicit requirement: "The story MUST have at least 2 player-facing dilemmas. A story with only 1 arc is a linear rail, not interactive fiction."
- Add explicit requirement: "Every path MUST include at least one beat AFTER its commit beat. The consequence beat shows the player what their choice led to."

### Acceptance Criteria

- [ ] Running SEED with a prompt that produces only 1 arc raises a hard error and blocks the pipeline
- [ ] Running SEED with a path missing a post-commit beat raises a hard error and blocks the pipeline
- [ ] Neither warning appears in `logs/debug.jsonl` for a valid run (they are now errors)
- [ ] The serialize prompt explicitly names both requirements

### Verification

```bash
# After fix: these patterns must not appear as WARNING level in valid runs
grep '"level": "WARNING"' projects/*/logs/debug.jsonl | grep -E "arc|scaffold"
# should return 0 matches for valid stories

# For invalid input, pipeline must exit non-zero and log ERROR
uv run qf run --to seed --project <linear-story-project>
echo "Exit code: $?"  # must be non-zero
```

pvliesdonk:
## Summary

- **Promotes post-commit beat check** from `SeedErrorCategory.WARNING` → `COMPLETENESS`: any path that has a commits beat but no subsequent consequence beat now blocks serialization and triggers LLM retry with actionable feedback
- **Converts arc count check** from `log.warning` to `raise SeedStageError`: a linear story (`arc_count < max(2, max_arcs // 4)`) cleanly fails before graph commit with a clear message
- **Strengthens serialize_seed_sections prompt**: adds MINIMUM BRANCHING REQUIREMENT and CONSEQUENCE BEAT REQUIREMENT blocks so the model knows these constraints upfront (prompt-engineer subagent reviewed)
- **Fixes GROW intersection phase**: prevents same-dilemma beats via interleave guard + 1-beat-per-dilemma candidate groups; strengthens `grow_phase3_intersections` prompt with explicit same-dilemma rule (Rule 6) and prefer-early-beats rule (Rule 7); adds predecessor conflict feedback to retry loop

## Test plan

- [x] All 193 unit tests pass (`uv run pytest tests/unit/test_mutations.py -q`)
- [x] 33 tests updated: consequence beats added to minimal fixtures, 3 warning-behavior tests updated to assert COMPLETENESS hard-error semantics
- [x] ruff format + mypy pass (pre-commit hooks)

## Design Conformance

Changes are enforcement of constraints already specified in the design (every path must have consequence beats; stories must have branching). No new schema fields or pipeline phases added — this is correctness enforcement of existing requirements.

Closes #1122

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -310,6 +310,7 @@ qf inspect -p <project> --json # Machine-readable JSON output
 - Using "tests pass" as sole evidence that a removal/refactoring issue is complete
 - Epics larger than 10 issues (split into sequential milestones)
 - Claiming "SEED/GROW/POLISH ran successfully" based on exit code without verifying output against design docs
+- **Silent degradation of story structure constraints** — if the pipeline cannot satisfy a structural requirement (cross-dilemma ordering, intersection formation, DAG consistency), it MUST fail loudly. Silently skipping constraints and producing a weaker story is not acceptable. `interleave_cycle_skipped` warnings, all-intersections-rejected, and similar "we tried but gave up" outcomes are pipeline failures, not warnings.
 
 ## Design Conformance Review (CRITICAL)
 
--- a/prompts/templates/grow_phase3_intersections.yaml
+++ b/prompts/templates/grow_phase3_intersections.yaml
@@ -28,8 +28,19 @@ system: |
      multiple candidate groups, choose the group where it fits most naturally
      and skip the others that use that beat.
   4. Do NOT force intersections. Rejecting a weak group is better than
-     accepting a forced one.
+     accepting a forced one. If no candidate group forms a natural scene,
+     return `{"intersections": []}`.
   5. Use ONLY beat IDs from the Valid IDs section. Do not invent or modify IDs.
+  6. CRITICAL: Each intersection MUST combine beats from DIFFERENT dilemmas.
+     The dilemma tag shown as [dilemma::X] after each beat ID tells you which
+     dilemma the beat belongs to. NEVER put two beats with the same dilemma
+     tag in the same intersection.
+     BAD: beat A [dilemma::trust] + beat B [dilemma::trust] + beat C [dilemma::flee]
+     GOOD: beat A [dilemma::trust] + beat B [dilemma::flee] + beat C [dilemma::search]
+  7. PREFER early beats (_beat_01, _beat_02). Avoid late beats (_beat_03, _beat_04)
+     because they often carry ordering dependencies on beats from other paths and
+     will fail validation. When in doubt, pick the earlier-numbered beat from each
+     dilemma in the candidate group.
 
   ## Candidate Groups
   {candidate_groups}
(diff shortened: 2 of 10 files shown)


pvliesdonk:
Addressed all review comments (commit 34a4ef2):

- **Fixed `_group_by_location` dedup logic** (🔴 Gemini + Claude): extracted `_limit_one_beat_per_dilemma()` helper using `isdisjoint`/`update(dilemmas)` — marks ALL of a beat's dilemmas as seen, prevents same-dilemma beats slipping through when a beat covers multiple dilemmas. Eliminates the `next(iter(set))` non-determinism.
- **Applied same limit to `_group_by_entity`** (🔴 Claude): entity-based candidates now also go through `_limit_one_beat_per_dilemma()`.
- **Moved arc count fail-fast before Phase 6** (🟡 Claude): `SeedStageError` check now fires immediately after pruning, before `serialize_dilemma_relationships` runs — avoids wasted LLM call on linear-story rejections.
- **Added test: same-dilemma temporal hint guard** (🟡 Claude): verifies that a `temporal_hint` with `relative_to == beat's own dilemma` produces no predecessor edges.
- **Added test: arc count `SeedStageError`** (🟡 Claude): verifies that `compute_arc_count` returning 1 raises `SeedStageError` matching "arc".

/gemini review

pvliesdonk:
Addressed Gemini's second-round comments (commit dd64d03):

- **Fixed `beat_id_to_dilemma` → `beat_id_to_dilemmas`**: Changed from `dict[str, str]` to `dict[str, set[str]]` (defaultdict(set)). Intersection beats that belong to paths from multiple dilemmas now accumulate all their dilemma IDs instead of the last one overwriting earlier ones.
- **Fixed same-dilemma guard**: Updated to `relative_to in beat_own_dilemmas` instead of equality with a single string — correctly catches same-dilemma hints for multi-dilemma beats.

/gemini review

PR #1267, opened 2026-04-18
docs(claude): make design specs authoritative over code and tests

pvliesdonk:
## Summary

Resolves an ambiguity in CLAUDE.md. The "Rules vs Guidelines" section said *"Design docs in `docs/design/` are guidelines"*, but later sections correctly called the three key doc sets (`how-branching-stories-work.md`, `story-graph-ontology.md`, `procedures/*.md`) *"authoritative"*. Those can't both be true.

This PR picks the correct answer: the three doc sets are **authoritative specifications**, and code and tests must conform to them — not the other way around.

## What changed

Three places in `CLAUDE.md`:

### 1. New subsection `Design Doc Authority (Specs Supersede Code)` (Instruction Hierarchy)

Lists the three authoritative doc sets and states four rules in practice:

1. Specs supersede code and tests. Conflicts mean the code/tests are wrong.
2. Docs-first fix order: read the spec; if it's silent or wrong, update the spec *first*, then fix the code. Never retroactively adjust a spec to describe what broken code does.
3. Surface drift explicitly — never silently paper over spec/code divergence.
4. ADRs and implementation notes cannot override authoritative specs.

Also carves out which docs are non-authoritative (CLAUDE.md itself, `01-prompt-compiler.md`, `07-getting-started.md`, `00-spec.md` which is deprecated).

### 2. Debugging Policy: added step 0

Read the authoritative spec first. Hypotheses about bugs should name the violated spec rule (e.g., *"code produces Y but SEED §R-3.6 requires Z"*). If the spec is silent, update the spec first.

### 3. Anti-Patterns to Avoid: two new entries

- Retroactively updating a spec to match broken code.
- Treating code or tests as the source of truth for pipeline behavior.

## Why now

Track 2 just finalized the three authoritative doc sets as LLM-debuggable specs (numbered rules, violations tables, stage-chaining contracts). The whole point of that work was to make the specs the primary reference — CLAUDE.md should state that plainly.

## Test plan

- [ ] Read the new section top-to-bottom; confirm it's unambiguous
- [ ] Check that it resolves the prior contradiction (line 30 vs line 302-304)
- [ ] Spot-check Debugging Policy step 0 and the two new Anti-Pattern entries for consistency with the new section

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -27,7 +27,33 @@ Context compaction loses nuance. Re-reading these rules prevents repeating mista
 | **Should/Prefer** | Strong defaults - deviate only with reason | "Should use targeted tests" |
 | **May/Can** | Options - use judgment | "May split into multiple PRs" |
 
-Design docs in `docs/design/` are guidelines. This CLAUDE.md file is rules.
+This CLAUDE.md file is rules for *how to work*. Separately, the authoritative design docs are rules for *what to build* — see the next section.
+
+### Design Doc Authority (Specs Supersede Code)
+
+The following design documents are **authoritative specifications**, not guidelines. They define what the pipeline must do; code and tests must conform to them.
+
+- `docs/design/how-branching-stories-work.md` — narrative model
+- `docs/design/story-graph-ontology.md` — graph / data model
+- `docs/design/procedures/*.md` — per-stage algorithm specifications
+
+**What "authoritative" means in practice:**
+
+1. **Specs supersede code and tests.** If code or tests conflict with these docs, **the code or tests are wrong.** Do not "interpret" the doc to match the code. Do not treat code as the source of truth.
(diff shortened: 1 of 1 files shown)


pvliesdonk:
Addressed all review feedback.

**Gemini:**
- Line 30: "see the next section" → "see §Design Doc Authority below" for consistent section-link format.
- "Explicitly marked authoritative" ambiguity: the authoritative set is the list in this section. Adding a new spec requires updating the list; a title or header inside a doc claiming "authoritative" does not count.

**Claude:**
- Docs-first fix order step ii: specifies the alignment mechanism — raise the proposed spec change in the current conversation, PR description, or a dedicated spec-update commit before touching code. Never fold a spec update into an unrelated implementation PR.
- Debugging step 0: softened "is evidence that code diverged" → "is usually evidence ... less often evidence of a test written against undefined behavior or a spec gap." In either case, the fix still starts with the spec.

Ready for re-review.

PR #1399, opened 2026-04-25
fix(fill): echo Literal allowed values in repair feedback (Cluster D-1)

### LINKED ISSUE #1398: [prompt-fix] Cluster D-1 — FILL repair-feedback enrichment
Phase 3 Cluster D-1 from the [prompt-vs-spec audit](docs/superpowers/reports/2026-04-25-prompt-spec-audit.md). Generalises the SEED Phase-1 `extra_repair_hints` pattern (PR #1384) to FILL `_build_error_feedback`.

## Problem

Per the audit FILL §`fill_phase0_voice.yaml` finding: when `VoiceDocument` validation fails on a Literal field (`pov` / `voice_register` / `sentence_rhythm`), the repair feedback names the field but does NOT echo the allowed values. A 4B model that outputs `pov: "third person limited"` sees only "value is not a valid enumeration member" with no list of what IS valid — repair-loop blindness.

## Fix

Extend `_build_error_feedback` in `src/questfoundry/pipeline/stages/fill.py` to:

1. For each `invalid_field` from `_classify_validation_error`, walk the Pydantic schema by dotted path and detect `Literal` types.
2. Append "Allowed values for `field.path`: a, b, c" lines to the feedback so the model sees the valid set on retry.
3. Behaviour-preserving: structural errors and non-Literal content errors are unchanged.

## Verification

- [ ] New tests pin the enriched feedback for `pov`, `voice_register`, and `sentence_rhythm` failures.
- [ ] Test confirms structural errors (missing field) don't trigger Literal lookups.
- [ ] Test confirms non-Literal content errors (e.g., `min_length=1` violation) don't crash.
- [ ] `uv run pytest tests/unit/test_fill*.py -x -q` passes.

pvliesdonk:
## Summary

Phase 3 Cluster D-1 from the [prompt-vs-spec audit](docs/superpowers/reports/2026-04-25-prompt-spec-audit.md). Closes #1398. Generalises the SEED Phase-1 `extra_repair_hints` pattern (PR #1384) to FILL `_build_error_feedback`.

When a Voice Document content failure fires on a Literal field (`pov`, `voice_register`, `sentence_rhythm`), the prior repair feedback named the field but did NOT echo the allowed values. A 4B model that emitted `pov: "third person limited"` saw "value is not a valid enumeration member" with no list of what IS valid — repair-loop blindness, the murder1 failure shape generalised to FILL.

## Changes

| File | Change |
|---|---|
| `src/questfoundry/pipeline/stages/fill.py` | New `_get_literal_values_at_path()` helper walks the schema by dotted path, returns Literal values (or None). `_build_error_feedback` now accepts `invalid_fields: list[str] \| None` and appends "Allowed values for `field.path`: ..." per-Literal field. Sole call site updated. |
| `tests/unit/test_fill_stage.py` | 4 new tests: pov + voice_register Literal echoing; structural failure skips lookup; non-Literal content failure (`min_length` violation) doesn't crash or add spurious lines. |

Diff: 2 files, +204/-1.

## Test plan

- [x] 140 FILL tests pass (`uv run pytest tests/unit/test_fill_stage.py tests/unit/test_fill_models.py -x -q`)
- [x] mypy + ruff clean on `src/questfoundry/pipeline/stages/fill.py`
- [x] Behaviour-preserving for non-Literal failures (`min_length`, missing fields, etc.)

## Out of scope

DRESS `_dress_llm_call` repair-feedback enrichment lands in **Cluster D-2** — separate stage code, different message format, independent review surface.

### DIFF
--- a/src/questfoundry/pipeline/stages/fill.py
+++ b/src/questfoundry/pipeline/stages/fill.py
@@ -778,7 +778,10 @@ class FillStage:
             for field_path in invalid_fields:
                 values = _get_literal_values_at_path(output_schema, field_path)
                 if values is not None:
-                    formatted = ", ".join(repr(v) for v in values)
+                    # Backtick-wrap each value for consistency with the SEED
+                    # repair-feedback pattern (serialize.py) and the prompts
+                    # themselves (CLAUDE.md §9 rule 1).
+                    formatted = ", ".join(f"`{v}`" for v in values)
                     literal_hints.append(f"Allowed values for `{field_path}`: {formatted}")
             if literal_hints:
                 parts.append("\n" + "\n".join(literal_hints))
--- a/src/questfoundry/pipeline/stages/fill.py
+++ b/src/questfoundry/pipeline/stages/fill.py
@@ -130,6 +130,76 @@ _STRUCTURAL_ERROR_TYPES = frozenset(
 )
 
 
+def _get_literal_values_at_path(
+    model_cls: type[BaseModel], field_path: str
+) -> tuple[Any, ...] | None:
+    """Walk a Pydantic model by dotted path and return the Literal values for that field.
+
+    Used by ``_build_error_feedback`` to echo allowed enum values on retry so
+    a small model that emits an out-of-set value (e.g. ``pov: "third person
+    limited"`` for the ``Literal["first_person", "second_person", ...]``
+    field) sees the valid set in the next attempt.
+
+    Args:
+        model_cls: The top-level Pydantic schema (e.g. FillPhase0Output).
+        field_path: Dotted path from `_classify_validation_error` (e.g.
+            ``"voice.pov"``).
+
+    Returns:
+        Tuple of allowed literal values if the field's annotation is a
+        ``Literal[...]`` (or a union containing one), else ``None``.
+    """
+    import types
+    import typing
+
+    # Strip list-index segments produced by Pydantic for list-typed errors
+    # (e.g. "passages.0.entity_updates.1.entity_id" -> "passages.entity_updates.entity_id"
+    # for schema traversal). Numeric segments map to "the item type", so we
+    # skip them and continue with the next named field.
+    parts = [p for p in field_path.split(".") if p and not p.isdigit()]
+
+    current_cls: type[BaseModel] | None = model_cls
+    annotation: Any = None
+    for i, part in enumerate(parts):
+        if current_cls is None:
+            return None
+        field_info = current_cls.model_fields.get(part)
+        if field_info is None:
+            return None
+        annotation = field_info.annotation
+
+        # Step into a nested BaseModel for the next segment, or into the item
+        # type of a list[BaseModel].
+        if i < len(parts) - 1:
+            inner: type[BaseModel] | None = None
+            args = getattr(annotation, "__args__", ())
+            if isinstance(annotation, type) and issubclass(annotation, BaseModel):
+                inner = annotation
+            else:
+                for arg in args:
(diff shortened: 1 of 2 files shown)


pvliesdonk:
Addressed all six items from the two review iterations in commit e3619a80:

| Item | Severity | Status |
|---|---|---|
| 1. Coverage gate failing at 65% (Optional[Literal[...]] union branches untested) | **Blocker** | ✅ Removed the speculative union branches per CLAUDE.md YAGNI — no current FILL field uses `Optional[Literal[...]]`. Helper is now ~25 lines instead of ~70; every retained branch is exercised by the new tests. |
| 2. Mixed structural+content failure: Literal hints surfaced during structural feedback | Moderate | ✅ Added `failure_type != "structural"` guard. New test `test_structural_failure_skips_literal_hints_even_with_invalid_fields` pins this with non-empty `invalid_fields`. |
| 3. Deferred `import types` / `import typing` | Minor | ✅ Hoisted to module-level (`from typing import ..., Literal, get_args, get_origin`). Helper uses the imported names directly. |
| 4. Test name misleading (`test_structural_failure_does_not_invoke_literal_lookup` actually tested `invalid_fields=None`) | Minor | ✅ Split into two: the renamed `test_structural_failure_skips_literal_hints_even_with_invalid_fields` (now matches the actual call-site shape) plus `test_none_invalid_fields_omits_literal_hints` for the legacy-caller default. |
| 5. `sentence_rhythm` not exercised | Minor | ✅ New `test_literal_field_failure_lists_allowed_sentence_rhythm_values`. |
| 6. List-index stripping had a comment but no test | Minor | ✅ New `test_get_literal_values_at_path_strips_list_index_segments` plus `test_get_literal_values_at_path_returns_none_for_missing_field` as direct unit tests of the helper. Also added the missing docstring on the voice_register test. |

Per the strengthened ~/.claude/CLAUDE.md rule — "low-priority follow-up" framings are treated as "do this now in this PR" — every item is folded in. 144 tests pass; mypy + ruff clean. Codecov should now hit 100% on the patch.

pvliesdonk:
Both follow-on items addressed in commit 781db56c:

| Item | Status |
|---|---|
| 1. `list[BaseModel]` inner traversal untested | ✅ New `test_get_literal_values_at_path_walks_list_of_basemodel` exercises the real-world case (`BatchedExpandOutput.blueprints` → `ExpandBlueprint.opening_move`) and asserts the four `opening_move` values are returned. |
| 2. Missing docstring on `test_literal_field_failure_lists_allowed_voice_register_values` | ✅ Added — matches the style of the pov and sentence_rhythm sibling tests. (My previous resolution comment claimed this was done but the edit didn't actually land — apologies for the noise.) |

The `list[BaseModel]` test catches a real coverage gap that would otherwise have shipped as silent repair-loop blindness on `blueprints.0.opening_move` failures, exactly the murder1 failure shape this PR is supposed to fix end-to-end.

pvliesdonk:
Fixed in commit 51f5e2ca — sharp catch.

The structural-guard test was vacuous as written: passing `invalid_fields=["passage.passage_id"]` (a `str` field, not `Literal`) meant `_get_literal_values_at_path` returned `None` regardless of the guard, so the test passed even if the guard were silently removed. Switched to `BatchedExpandOutput` + `"blueprints.0.opening_move"` (a real `Literal` field), so the helper returns the four valid values and only the guard suppresses the hint. Verified independently: removing the guard now causes the test to fail correctly.

Docstring also updated to document why this specific schema/path was chosen — future-me would otherwise be tempted to swap it for a simpler model.

PR #1441, opened 2026-04-26
fix(seed): enforce single-path → hard classification in dilemma_analyses (Cluster F-12)

pvliesdonk:
## Summary

**Blocked on PR #1443** (spec rule must land first per CLAUDE.md §Design Doc Authority). This PR was rewritten around the narrower R-7.6 framing once the spec direction was course-corrected.

Closes #1439. Surfaced by investigation of the \`projects/murder2/\` run failure: SEED produced \`dilemma::locket_planted_or_dropped\` with \`dilemma_role: "soft"\`, \`explored: ["planted"]\`, **\`convergence_point: "paths converge during final accusation scene"\`**. GROW R-6.4 caught the structural error correctly (no sibling path to merge from), but the SEED-side prompt should not have produced a non-null convergence_point for a single-path dilemma.

## Why the narrower rule (corrected from first draft)

The first draft of F-12 forced single-path dilemmas to \`dilemma_role: "hard"\`. **That was a category error.** \`dilemma_role\` is a property of the answer pair (does it create incompatible world states?), not of how many paths happened to be developed. The F-7 cluster's \`dilemmas_prompt\` already documents single-path soft as a legitimate "flavor" pattern.

The narrower rule preserves the conceptual classification and only constrains the structural pointer GROW reads:

- \`dilemma_role\` stays as classified by stakes — \`soft\` for naturally-soft dilemmas even when only one answer was explored.
- \`convergence_point\` MUST be null when \`explored\` has only one answer (R-7.6) — there's nothing to converge with.

## Changes

In \`prompts/templates/serialize_seed_sections.yaml dilemma_analyses_prompt\`:

1. **Replace "Structural override forces hard"** with "Single-path constraint (SEED R-7.6)": preserves role, gates \`convergence_point\` only.
2. **Restore Decision Process to 2 passes** (no Pass 0 forcing hard) plus a new "Convergence Point pass" applied per-dilemma after classification.
3. **Replace Single-Path Examples**: GOOD shows single-path soft with \`convergence_point: null\`; BAD shows single-path soft with non-null convergence_point.
4. **Update Convergence Point section**: explicit triple rule (hard → null; single-path → null per R-7.6; soft+both-explored → location-based).
5. **Update Rules**:
   - "Path count does NOT change the role (it only gates \`convergence_point\` per R-7.6)"
   - New guard: "Do NOT downgrade a conceptually-soft dilemma to hard just because only one answer was explored — keep the role and set \`convergence_point: null\`."
[...]

Issue #1439, opened 2026-04-26
[prompt-fix] SEED dilemma_analyses_prompt: enforce single-path → hard classification (murder2 failure)


pvliesdonk:
Surfaced by investigation of the \`projects/murder2/\` run (failed at GROW convergence validation, 95s in).

## Failure

GROW halted with a structural violation: \`dilemma::locket_planted_or_dropped\` was classified by SEED as \`dilemma_role: "soft"\` while only one path (\`planted\`) was explored — leaving no sibling path for the convergence beat to merge with.

GROW's R-7.4 invariant requires every soft dilemma to have a structural convergence beat where paths rejoin. With only one path generated for the locket dilemma, that beat cannot exist.

## Root cause (prompt gap)

\`prompts/templates/serialize_seed_sections.yaml\` \`dilemma_analyses_prompt\` (the Self-Check section that classifies dilemmas as \`hard\` / \`soft\`) was tightened in F-7 (PR #1432) to remove the legacy \`flavor\` role and enforce \`hard\`-or-\`soft\`-only output. But it does **not** enforce a path-count → role constraint:

- A **soft** dilemma classification commits to a downstream convergence beat. That requires BOTH paths to exist in the SEED output (one shared sibling for the soft path to converge into).
- A **hard** dilemma classification means paths run to separate endings and never meet — single-path is fine because the unexplored answer is a shadow.

So the classification rule should be: **if a dilemma's \`explored\` list contains only ONE answer, the dilemma MUST be classified \`hard\`** (no convergence beat to author). Soft is reserved for dilemmas with both answers explored.

## Spec citation

- \`docs/design/procedures/seed.md §R-7.1\` — \`dilemma_role\` ∈ {\`hard\`, \`soft\`} (already enforced by F-7)
- \`docs/design/procedures/grow.md §R-7.4\` — soft dilemmas require a convergence beat (this is what fired in murder2)
- \`docs/design/procedures/seed.md §R-5.1\` — at least 2 dilemmas must have both answers explored (covered by F-7's MINIMUM BRANCHING block in \`discuss_seed.yaml\`)

## Fix sketch

In \`serialize_seed_sections.yaml\` \`dilemma_analyses_prompt\`:

1. Add a **path-count gate** at the top of the classification rule:
[...]

PR #1443, opened 2026-04-26
spec(seed): add R-7.6 — single-path Dilemma's convergence_point MUST be null

### LINKED ISSUE #1442: [spec] SEED §Phase 7: codify single-path → hard rule (R-7.6 candidate)
Surfaced by review of #1441 (Cluster F-12 — single-path → hard classification).

## Problem

PR #1441 added a structural-override rule to \`serialize_seed_sections.yaml dilemma_analyses_prompt\`: if a dilemma's \`explored\` list contains only ONE answer, it MUST be classified \`hard\`. This was driven by the murder2 run failure (GROW R-6.4 halted on a single-path dilemma classified \`soft\`).

The prompt now encodes this as a generation rule, but \`docs/design/procedures/seed.md §Phase 7\` (the dilemma classification section) does not codify it as a SEED-side invariant. The existing SEED rule numbers (R-7.1 through R-7.5) cover hard/soft enum membership, residue/ending independence axes, and so on — none cover the path-count constraint.

Per CLAUDE.md §Design Doc Authority: docs-first order. Encoding a new invariant in a prompt without anchoring it in the spec leaves the spec lagging the prompt — exactly the drift the docs-first policy is designed to prevent.

## Proposed change

Add a new rule under \`docs/design/procedures/seed.md §Phase 7 §Rules\`:

\`\`\`
R-7.6. A Dilemma whose \`explored\` list contains only one Answer MUST be
classified with \`dilemma_role: "hard"\` and \`convergence_point: null\`.
A single-path \`soft\` classification will halt GROW at Phase 6 R-6.4
validation (no sibling path for the convergence beat to merge with).
\`\`\`

And add a Violations table entry:

| Symptom | Root cause | Broken rule |
|---|---|---|
| Dilemma has \`explored: [a]\`, \`unexplored: [b]\`, \`dilemma_role: "soft"\` | Path-count constraint not applied at classification time | R-7.6 (caught downstream by GROW R-6.4) |

## Verification

- [ ] \`seed.md §Phase 7\` has new R-7.6 rule
- [ ] Violations table includes the murder2-style symptom
- [ ] \`serialize_seed_sections.yaml dilemma_analyses_prompt\` citations switch from \`GROW R-6.4\` to \`SEED R-7.6 (downstream-validated by GROW R-6.4)\` for clarity (optional follow-up)

## Related

- F-12 (#1439, PR #1441) — prompt-side fix that landed first
- murder2 run failure (\`projects/murder2/logs/debug.jsonl\`) — original trigger
- GROW R-6.4 (\`docs/design/procedures/grow.md:448\`) — the downstream validator that fires when this is violated

pvliesdonk:
## Summary

Closes #1442. Codifies the structural invariant surfaced by the \`projects/murder2/\` run failure: when a Dilemma's \`explored\` list contains only one Answer, \`convergence_point\` MUST be null regardless of \`dilemma_role\`.

**Per CLAUDE.md §Design Doc Authority — "spec first, always":** PR #1441 (Cluster F-12 prompt fix) is **blocked on this PR**. Once R-7.6 lands, F-12 rewrites around the narrower rule.

## Why this rule (not the wrong-direction first draft)

The first draft of R-7.6 forced single-path dilemmas to \`dilemma_role: "hard"\`. **That was a category error**: \`dilemma_role\` is a property of the QUESTION/STAKES (does this answer pair create incompatible world states?), not of how many paths happened to be developed. The F-7 cluster's own \`dilemmas_prompt\` spells this out:

> \"'Soft' dilemmas (only the default explored) add flavor without adding branching arcs.\"

Forcing such a dilemma to \`hard\` would break the conceptual classification to suit the structural validator.

The narrower rule preserves the conceptual classification and only constrains the structural pointer GROW reads:

- \`dilemma_role\` stays \`soft\` for a "naturally soft" dilemma whose alternate answer was deliberately left as a shadow.
- \`convergence_point\` MUST be null because there is no second path to converge with.
- GROW R-6.4 only fires when \`convergence_point\` is non-null and the corresponding beat cannot be located; a null \`convergence_point\` skips the check, so single-path soft dilemmas pass GROW without a structural error.

## Trigger

\`projects/murder2/\` graph.db had:

\`\`\`
"locket_planted_or_dropped": {
  "dilemma_role": "soft",                                     # correct (semantic)
  "explored": ["planted"],
  "unexplored": ["dropped"],
  "convergence_point": "paths converge during final accusation scene"   # WRONG
}
\`\`\`

The bug is in the last field — SEED set a non-null \`convergence_point\` for a single-path dilemma. GROW R-6.4 caught it correctly, but the SEED-side prompt should not have produced it.

## Changes (4 lines, 1 file)

1. **New \`R-7.6\` rule** in \`docs/design/procedures/seed.md §Phase 7 §Rules\`.
2. **New entry in §Phase 7 Violations table** for the murder2-style symptom.
3. **New entry in §Rule Index footer**.

## Spec citations

- \`docs/design/procedures/seed.md §Phase 7\` — Dilemma Analysis (this PR adds R-7.6)
- \`docs/design/procedures/grow.md §R-6.4\` (line 448) — downstream convergence validator
- \`CLAUDE.md §Design Doc Authority\` — spec first, always

## Related

- #1439 — original prompt-fix tracker (will be retargeted to the narrower rule once spec lands)
- #1442 — this spec gap (closed by this PR)
- PR #1441 (Cluster F-12) — blocked on this PR; will rewrite around the narrower rule after merge

### DIFF
--- a/docs/design/procedures/seed.md
+++ b/docs/design/procedures/seed.md
@@ -422,6 +422,8 @@ R-7.4. `residue_weight` and `dilemma_role` are independent axes — a soft dilem
 
 R-7.5. If the LLM call fails, defaults are applied (role: soft, weight: light, salience: low) but the failure is logged at WARNING level with the Dilemma IDs affected. Silent default application is forbidden.
 
+R-7.6. When a Dilemma's `explored` list (from Phase 2) contains only one Answer, `convergence_point` MUST be `null` regardless of `dilemma_role`. With only one path, there is no structural convergence to point to. The conceptual classification (`hard` vs `soft`) is preserved — `soft` remains correct for a dilemma whose alternate answer was deliberately left as a shadow per the F-7 flavor pattern (only the default explored, single arc) — but the structural pointer is not applicable. GROW Phase 6 R-6.4 ("soft Dilemma without structural convergence → halt") fires only when `convergence_point` is non-null and the corresponding beat cannot be located; a null `convergence_point` skips that check, so single-path soft dilemmas pass GROW without a structural error.
+
 **Violations:**
 
 | Symptom | Root cause | Broken rule |
@@ -429,6 +431,7 @@ R-7.5. If the LLM call fails, defaults are applied (role: soft, weight: light, s
 | Dilemma has no `dilemma_role` field after Phase 7 | Phase 7 soft-failed and left the field unset | R-7.1 / R-7.5 |
 | Dilemma has `dilemma_role: flavor` | `flavor` is not a valid role (flavor-level choices are handled by POLISH false branches, not by dilemma role) | R-7.1 |
 | Phase 7 LLM failure produces no log entry | Silent default application | R-7.5 |
+| Dilemma has `explored: [a]`, `unexplored: [b]`, non-null `convergence_point` | Convergence pointer set despite no second path to converge with; will halt GROW R-6.4 unable to locate the beat | R-7.6 (downstream-validated by GROW R-6.4) |
 
 ### Output Contract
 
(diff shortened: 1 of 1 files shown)


pvliesdonk:
Major disapproval:
A dilemma cannot be *hard* just because there is only one path explored. That makes no semantic or narrative sense 
The actual problem is deeper.

pvliesdonk:
Closing — both layering attempts (force-hard / force-convergence_point-null) were wrong. The murder2 failure is a GROW Phase 6 R-6.4 wording gap: the Operations header (grow.md:438) scopes the operation to 'each Dilemma with dilemma_role: soft and two explored paths', but R-6.4's halt wording (grow.md:448) doesn't carry that scope. The fix is to either (a) scope R-6.4 explicitly to two-path soft, or (b) add an exemption rule for single-path soft.

#1442 stays open and will be closed by the GROW spec PR.

This branch is preserved for archaeology but should not be re-opened.

PR #1458, opened 2026-04-27
fix(polish): residue passages inherit entities from target passage (#1457)

### LINKED ISSUE #1457: POLISH: residue passages created without 'entities' field break FILL phantom-ID prevention
## Summary

`_create_residue_beat_and_passage` creates residue passage nodes without an \`entities\` field. FILL Phase 1 then runs the prose + extract calls with **no Valid Entity IDs context**, so the LLM hallucinates short-form entity IDs (e.g. \`clara\` instead of \`clara_yu\`), and \`_resolve_entity_id\` rejects them with hard escalations.

## Reproducer

\`projects/murder3\` POLISH→FILL run failed with 4 \`entity_update_missing_entity\` escalations on a single residue passage:

\`\`\`
passage::residue_clara_victim_or_accomplice__victim_beat_02_state_flag__clara_stolen_genius_committed
  field=appearance,  provided=clara
  field=appearance,  provided=clara
  field=mannerism,   provided=clara
  field=sensory,     provided=clara
\`\`\`

The actual entity in the graph is \`character::clara_yu\` (Clara Yu). The dilemma subject \`clara_victim_or_accomplice\` and consequence \`clara_stolen_genius\` both contain \`clara\` as a *prefix*, which the LLM mistook for the entity ID once it had no Valid Entity IDs section to anchor against. The source passage \`passage::clara_victim_or_accomplice__victim_beat_02\` has \`entities=['character::clara_yu', 'character::alistair_vance']\` — the residue inherits nothing.

## Scope

**All 28** residue passages in murder3 lack the \`entities\` field. Only 1 surfaced as a hard failure this run; the other 27 likely produced subtly wrong prose (no entity_updates emitted, or the LLM happened to produce legal IDs by luck) without crossing the \`_resolve_entity_id\` failure threshold. The bug is structural — every residue created since the residue-passage-with-variants strategy landed has missed inheriting entities.

## Root cause

\`src/questfoundry/pipeline/stages/polish/deterministic.py:1207-1271\` (\`_create_residue_beat_and_passage\`):

\`\`\`python
graph.create_node(
    residue_passage_id,
    {
        \"type\": \"passage\",
        \"raw_id\": f\"residue_{residue_suffix}\",
        \"summary\": rspec.content_hint or f\"Residue for {rspec.flag}\",
        \"requires\": [rspec.flag],
        \"is_residue\": True,
        \"residue_for\": rspec.target_passage_id,
        \"mapping_strategy\": rspec.mapping_strategy,
    },
)
\`\`\`

No \`entities\` key. The companion residue beat is created with \`entities: []\` (explicit empty list) — also wrong but less visible.

Downstream consequences in FILL Phase 1:

1. \`format_entity_states(graph, passage_id)\` reads \`passage.get(\"entities\", [])\` → \`[]\` → returns \`\"(no entities)\"\` for the prose context.
2. \`_fill_extract_call\` builds \`valid_entity_ids = [...]\` from \`passage.get(\"entities\", [])\` → \`[]\` → injects \`\"(none)\"\` into the Valid Entity IDs prompt slot.
3. LLM generates prose mentioning the source passage's characters by name and emits \`entity_updates\` with whatever short ID it can construct from the passage_id text.
4. \`_resolve_entity_id\` tries every category prefix on the hallucinated ID, finds no match, escalates \`missing_entity\`, and FILL halts at the contract gate.

## Fix

When creating the residue passage and beat, look up \`rspec.target_passage_id\` and copy its \`entities\` into both new nodes. The residue is narratively a continuation of the source — same cast.

\`\`\`python
target_passage = graph.get_node(rspec.target_passage_id) or {}
target_entities = list(target_passage.get(\"entities\") or [])

# beat:
graph.create_node(beat_id, {..., \"entities\": target_entities, ...})

# passage:
graph.create_node(residue_passage_id, {..., \"entities\": target_entities, ...})
\`\`\`

## Out of scope

- **Defense in depth in FILL**: \`_fill_extract_call\` could fall back to all-graph-entities when \`valid_entity_ids\` is empty. Useful but separate — empty entities is a structural problem the prompt shouldn't paper over. File separately if needed.
- **The other 27 silently-wrong residue passages in murder3**: their prose may already be correct (luck) or subtly wrong. The fix here prevents the bug going forward; existing prose can be re-run if the user cares (POLISH rewinds residue passages).
- **\`mapping_strategy=parallel_passages\`**: same pattern likely applies once that strategy is implemented (per #1262 / docs/superpowers/specs/2026-04-20-polish-compliance-design.md). Not implemented today, so out of scope.

## Verification

\`\`\`python
# After fix:
assert residue_passage[\"entities\"] == source_passage[\"entities\"]
\`\`\`

\`\`\`bash
$ uv run pytest tests/unit/test_polish_deterministic.py -k residue -x -q
# new test: \"residue passage inherits entities from target passage\" passes
\`\`\`

pvliesdonk:
## Summary

Closes #1457.

Residue passages (and their companion beats) were created without an \`entities\` field. FILL Phase 1 builds its \`### Valid Entity IDs\` prompt context from \`passage[\"entities\"]\`, so the slot became \`(none)\` and the prose-call LLM hallucinated short-form entity IDs from passage_id text — \`clara\` instead of \`character::clara_yu\`. \`_resolve_entity_id\` then escalated \`missing_entity\` and FILL halted at the contract gate.

## Reproducer

\`projects/murder3\` POLISH→FILL run halted on:

\`\`\`
passage::residue_clara_victim_or_accomplice__victim_beat_02_state_flag__clara_stolen_genius_committed
  field=appearance,  provided=clara
  field=appearance,  provided=clara
  field=mannerism,   provided=clara
  field=sensory,     provided=clara
\`\`\`

The source passage \`passage::clara_victim_or_accomplice__victim_beat_02\` has \`entities=['character::clara_yu', 'character::alistair_vance']\`. The residue inherited nothing.

## Scope

**All 28** residue passages in murder3 lack the field. Only 1 surfaced as a hard failure this run; the rest produced silently-wrong prose (no entity_updates emitted, or phantom IDs that happened to resolve). The bug is structural — every residue created since the residue layer landed has missed inheriting entities.

## Fix

| Change | What |
|---|---|
| New helper \`_residue_inherited_entities(graph, target_passage_id)\` | Reads target's \`entities\`, defensive against missing target / None / missing field. |
| \`_apply_residue_with_variants\` | Passes inherited list into BOTH the residue beat and residue passage \`create_node\` calls. |
| \`_apply_residue_parallel_passages\` | Same fix for the alternate strategy. |

## Out of scope

- **Defense in depth in FILL**: \`_fill_extract_call\` could fall back to all-graph-entities when \`valid_entity_ids\` is empty. Useful but separate — empty entities is a structural problem the prompt shouldn't paper over.
- **Re-running murder3 with corrected residues**: existing graph data won't backfill — POLISH would need to be rewound past the residue phase to regenerate. Out of scope for this PR.

## Tests

\`tests/unit/test_polish_apply.py::TestCreateResidueBeatAndPassage\` adds 3 tests:

- \`test_residue_with_variants_inherits_target_entities\` — primary regression
- \`test_residue_parallel_passages_inherits_target_entities\` — same contract for the alt strategy
- \`test_residue_inheritance_defensive_when_target_has_no_entities\` — graceful degrade to \`[]\`

\`\`\`
$ uv run pytest tests/unit/ -k polish -x -q
367 passed
\`\`\`

## Test plan

- [x] All polish unit tests pass (367)

### DIFF
--- a/docs/design/procedures/polish.md
+++ b/docs/design/procedures/polish.md
@@ -526,6 +526,8 @@ R-6.3. Any step failure rolls back the transaction. No graph mutations are commi
 
 R-6.4. Phase 6 creates new beat nodes only for residue beats and false-branch beats — never new narrative beats.
 
+R-6.5. Residue passages and their companion residue beats inherit the `entities` field from their target passage (the passage referenced by `residue_for`). The residue plays out at the same dramatic moment as its target — same cast on stage. FILL builds its `### Valid Entity IDs` prompt context from `passage["entities"]`; an empty or missing list there forces phantom-ID emissions in the prose call. Both `residue_passage_with_variants` and `parallel_passages` mapping strategies must apply this inheritance.
+
 **Violations:**
 
 | Symptom | Root cause | Broken rule |
@@ -533,6 +535,7 @@ R-6.4. Phase 6 creates new beat nodes only for residue beats and false-branch be
 | Half-built passage layer visible to Phase 7 | Atomicity violated | R-6.1 / R-6.3 |
 | Phase 6 creates a commit beat | Narrative-beat creation forbidden post-SEED | R-6.4 |
 | Phase 6 mutates an existing `belongs_to` edge | Should only add passage-layer and structural-beat nodes | R-6.4 |
+| Residue passage emits phantom entity IDs (e.g. `clara` instead of `character::clara_yu`) at FILL | Residue created with empty `entities` instead of inheriting from target | R-6.5 |
 
 ### Output Contract
 
--- a/src/questfoundry/pipeline/stages/polish/deterministic.py
+++ b/src/questfoundry/pipeline/stages/polish/deterministic.py
@@ -1213,15 +1213,26 @@ def _residue_inherited_entities(graph: Graph, target_passage_id: str) -> list[st
     on the residue forces the prose-call to invent IDs from the passage_id
     text and produces phantom-ID escalations downstream (#1457).
 
-    Defensive against the target passage being absent (resilient to call-order
-    issues) or its ``entities`` field being None / missing — both degrade to
-    an empty list rather than raising. The explicit ``is None`` check guards
-    the missing-target case; ``or []`` on the inner lookup handles a present
-    target with no entities or a None value.
+    A target passage that doesn't exist is a structural failure: phase 6's
+    application order (R-6.2) creates target passages before residues, so
+    this can only happen if Phase 4-5 produced a plan that violates that
+    order or someone called this helper out-of-band. Raise rather than
+    fall back to ``[]`` — silently inheriting an empty entity list is the
+    exact failure mode that produced #1457 in the first place. (Per
+    ``.gemini/styleguide.md`` anti-pattern: silent fallbacks that hide
+    bugs prefer explicit errors.)
+
+    The inner ``or []`` on ``target.get("entities")`` handles a *present*
+    target whose entities field is missing or explicitly ``None`` — that's
+    a legitimate empty case (e.g. setup beats before any cast is on stage).
     """
     target = graph.get_node(target_passage_id)
     if target is None:
-        return []
+        raise ValueError(
+            f"R-6.5: residue target passage {target_passage_id!r} not found in graph. "
+            "Phase 6 application order (R-6.2) requires target passages to be created "
+            "before residues."
+        )
     return list(target.get("entities") or [])
 
 
--- a/src/questfoundry/pipeline/stages/polish/deterministic.py
+++ b/src/questfoundry/pipeline/stages/polish/deterministic.py
@@ -1215,9 +1215,13 @@ def _residue_inherited_entities(graph: Graph, target_passage_id: str) -> list[st
 
     Defensive against the target passage being absent (resilient to call-order
     issues) or its ``entities`` field being None / missing — both degrade to
-    an empty list rather than raising.
+    an empty list rather than raising. The explicit ``is None`` check guards
+    the missing-target case; ``or []`` on the inner lookup handles a present
+    target with no entities or a None value.
     """
-    target = graph.get_node(target_passage_id) or {}
+    target = graph.get_node(target_passage_id)
+    if target is None:
+        return []
     return list(target.get("entities") or [])
 
 
--- a/src/questfoundry/pipeline/stages/polish/deterministic.py
+++ b/src/questfoundry/pipeline/stages/polish/deterministic.py
@@ -1204,6 +1204,23 @@ def _create_variant_passage(graph: Graph, vspec: VariantSpec) -> None:
     graph.add_edge("variant_of", vspec.variant_id, vspec.base_passage_id)
 
 
+def _residue_inherited_entities(graph: Graph, target_passage_id: str) -> list[str]:
+    """Return the entity IDs a residue should inherit from its target passage.
+
+    A residue passage / beat plays out narratively at the same dramatic moment
+    as its target — same cast on stage. FILL builds its `### Valid Entity IDs`
+    context from the passage's ``entities`` field, so a missing or empty list
+    on the residue forces the prose-call to invent IDs from the passage_id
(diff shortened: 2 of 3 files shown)


pvliesdonk:
Addressed claude-review notes in [c677de4](https://github.com/pvliesdonk/questfoundry/commit/c677de40):

- **Redundant `list()` wrappers** — removed at all four call sites; the helper already returns a fresh `list(...)`, and node data is JSON-serialised by `sqlite_store.set_node()` on store, so reads each get an independently-deserialised list. No aliasing footgun.
- **Spec gap** — added **R-6.5** to `polish.md` Phase 6: residue passages and their companion residue beats MUST inherit `entities` from their target passage. Documents the invariant the code now enforces, with a violation row pointing at the murder3 phantom-ID symptom.
- **Variant-passages companion gap** — filed as **#1459** rather than bundling. The investigation step (does FILL reach variants? is `entities=[]` reliably empty?) is non-trivial enough to warrant its own PR.

Re-running gemini below for a fresh review on the latest commit.

pvliesdonk:
Addressed gemini-code-assist's nit in [d13e922](https://github.com/pvliesdonk/questfoundry/commit/d13e9221): replaced `graph.get_node(...) or {}` with an explicit `if target is None: return []` guard in `_residue_inherited_entities`.

Notes:
- Kept the inner `target.get("entities") or []` rather than `target.get("entities", [])` — the former returns `[]` for both a missing key AND an explicit `None` value; gemini's suggested form would return `None` unchanged for a None value, then `list(None)` would `TypeError`. Slightly more defensive without behavioural change in well-formed graphs.
- The cited convention ("use explicit `is not None`") didn't match what the repo actually does — `graph.get_node(...) or {}` appears in ~17 places under `src/` (e.g. `pipeline/size.py:245`, `fill.py:1262/1442/1532/2106`, `dress.py:1455/1706/2006`, `fill_context.py:774/818/819/1062`, `polish_context.py:246`, `polish_validation.py:371/484`, `dress_context.py:393`). This PR doesn't sweep them — keeping scope tight. If you want a project-wide convention sweep separately, happy to file an issue.

`uv run pytest tests/unit/test_polish_apply.py -k residue -x -q` → 7 passed.

Re-running gemini below for a fresh review on the latest commit.

pvliesdonk:
Addressed gemini's `Silent fallbacks that hide bugs` finding in [4e624f7](https://github.com/pvliesdonk/questfoundry/commit/4e624f76):

- Replaced `if target is None: return []` with `raise ValueError(...)` naming R-6.5 + R-6.2. That early-return was itself the anti-pattern from `.gemini/styleguide.md` §4 — a residue with an empty entities list is exactly the failure mode that produced #1457. Phase 6's R-6.2 application order guarantees target passages exist before residues; a missing target is a structural pipeline failure, not a normal case.
- Updated `test_residue_inherited_entities_raises_for_absent_target` to assert the raise instead of the empty-list return.
- Kept `target.get("entities") or []` on the inner lookup — a present target with a missing/None entities field is a legitimate empty case (e.g. setup beats with no cast yet).

**Two other gemini nits not addressed:**

1. **Line 1225 (shared list aliasing)**: `sqlite_store.set_node` calls `json.dumps(data)` (line 305), so each `create_node` independently serializes its data dict at store time and `get_node` returns a freshly deserialized dict via `json.loads` on read. The two new nodes share a Python list reference only transiently in the same frame between `create_node` calls — there's no aliasing after storage. claude-review's earlier analysis is correct.

2. **Line 1261 (extract context object)**: `_apply_residue_with_variants` and `_apply_residue_parallel_passages` share exactly one line of setup (`inherited_entities = ...`). Adding a NamedTuple/dataclass context for one variable is over-engineering per CLAUDE.md "Don't add features, refactor, or introduce abstractions beyond what the task requires."

`uv run pytest tests/unit/test_polish_apply.py -k residue -x -q` → 8 passed.

Re-running gemini below for a fresh review on the latest commit.

PR #1460, opened 2026-04-27
fix(polish): variant passages inherit entities from base (#1459)

### LINKED ISSUE #1459: POLISH: variant passages likely share the residue-entities bug (#1457 companion)
## Summary

claude-review on PR #1458 flagged that \`_create_variant_passage\` (\`src/questfoundry/pipeline/stages/polish/deterministic.py:1191\`) creates variant passages without an \`entities\` field — exactly the same shape as the residue-passages bug fixed in #1457:

\`\`\`python
graph.create_node(
    vspec.variant_id,
    {
        \"type\": \"passage\",
        \"raw_id\": vspec.variant_id.split(\"::\")[-1],
        \"summary\": vspec.summary,
        \"requires\": vspec.requires,
        \"is_variant\": True,
    },
)

graph.add_edge(\"variant_of\", vspec.variant_id, vspec.base_passage_id)
\`\`\`

Variant passages sit at the same dramatic moment as their base passage. If FILL hits a variant through the same prose-call code path it uses for residue passages, the same phantom-ID failure can occur — \`passage[\"entities\"]\` is empty, the Valid Entity IDs prompt slot becomes \`(none)\`, and the LLM hallucinates short-form entity IDs from passage_id text.

## Confirmation needed

Before fixing, verify:

1. Does FILL Phase 1 generate prose for variant passages? (i.e. do they hit \`_fill_prose_call\` / \`_fill_extract_call\`?)
2. If yes, is \`entities=[]\` reliably empty on every variant passage in real graphs? (\`projects/murder3\` likely has the data — check post-POLISH.)

If both confirmed, the fix mirrors #1457 / R-6.5: read \`base_passage_id\`'s \`entities\` and pass into the variant \`create_node\`. Same defensive helper signature.

## Spec coverage

\`docs/design/procedures/polish.md\` R-6.5 (added in PR #1458) currently scopes only residue passages. If the variant-passage bug is confirmed, the rule should be extended (or a sibling rule added) for variants inheriting from \`variant_of\`.

## Reproducer

Likely \`projects/murder3\` — that graph has variant passages alongside the 28 residues. Pre-fix it would have escalated like the residue case if FILL had reached a variant. The murder3 FILL run halted on the residue side before any variant came up, so this is unconfirmed in production but structurally probable.

## Reference

- PR #1458 review comment (claude-review): https://github.com/pvliesdonk/questfoundry/pull/1458
- Sibling fix: #1457 / R-6.5 in polish.md

pvliesdonk:
## Summary

Closes #1459. Companion to #1457 / PR #1458.

\`_create_variant_passage\` created variant nodes without an \`entities\` field — same shape as the residue bug fixed in #1458. FILL reaches variant passages via the catch-all sweep at the end of \`_get_generation_order\` (\`fill.py:1120-1124\`), so the same phantom-ID failure mode applies once a project's POLISH plan produces variants.

## Reproducer

No in-the-wild repro: \`projects/murder3\` happens to have **0** variants (its POLISH plan didn't propose any). But the bug is structural — confirmed by reading the code paths:

1. \`_create_variant_passage\` (\`deterministic.py:1191\` pre-fix): no \`entities\` key set on the variant node.
2. \`_get_generation_order\` (\`fill.py:1118-1124\`): sweeps every passage with no prose, including variants.
3. FILL Phase 1 prose call reads \`passage.get(\"entities\", [])\` → \`[]\` → Valid Entity IDs slot becomes \`(none)\` → LLM hallucinates short-form IDs from passage_id text → \`_resolve_entity_id\` rejects → \`missing_entity\` escalation. Identical chain to #1457.

## Fix

Mirror of #1458 / R-6.5:

- Rename helper `_residue_inherited_entities` → `_inherited_passage_entities` since the invariant is now shared by residues and variants.
- `_create_variant_passage` reads \`vspec.base_passage_id\`'s entities and copies them into the new variant node.
- Same \`is None: raise ValueError\` shape per \`.gemini/styleguide.md\` anti-pattern (silent fallbacks that hide bugs).

## Spec update — R-6.5 extension

\`docs/design/procedures/polish.md\` R-6.5 now covers both residue and variant inheritance under a single rule (same underlying invariant: synthesized passage at the same dramatic moment as a source passage MUST inherit that source's entities). Two new violation-table rows: variant phantom-ID symptom, and the missing-source raise.

## Tests

| Test | What it pins |
|---|---|
| \`test_variant_inherits_entities_from_base_passage\` (new) | Primary regression: variant copies entities list from base |
| \`test_inherited_passage_entities_raises_for_absent_source\` (renamed from residue-specific) | Helper raises \`ValueError\` naming R-6.5 + R-6.2 — covers both residue and variant code paths |

Existing residue tests + the \`test_apply_full_plan_to_graph\` integration test continue to pass — the rename is a pure refactor at the call sites.

\`\`\`
$ uv run pytest tests/unit/ -k polish -x -q
369 passed
\`\`\`

## Test plan

- [x] All polish unit tests pass (369)

### DIFF
--- a/docs/design/procedures/polish.md
+++ b/docs/design/procedures/polish.md
@@ -526,7 +526,12 @@ R-6.3. Any step failure rolls back the transaction. No graph mutations are commi
 
 R-6.4. Phase 6 creates new beat nodes only for residue beats and false-branch beats — never new narrative beats.
 
-R-6.5. Residue passages and their companion residue beats inherit the `entities` field from their target passage (the passage referenced by `residue_for`). The residue plays out at the same dramatic moment as its target — same cast on stage. FILL builds its `### Valid Entity IDs` prompt context from `passage["entities"]`; an empty or missing list there forces phantom-ID emissions in the prose call. Both `residue_passage_with_variants` and `parallel_passages` mapping strategies must apply this inheritance.
+R-6.5. Synthesized passages inherit the `entities` field from their source passage on the beat layer. This applies to:
+
+- **Residue passages** and their companion **residue beats** — inherit from the passage referenced by `residue_for`. Both `residue_passage_with_variants` and `parallel_passages` mapping strategies must apply the inheritance.
+- **Variant passages** — inherit from the passage referenced by `variant_of` (the base passage).
+
+The synthesized node plays out at the same dramatic moment as its source — same cast on stage. FILL builds its `### Valid Entity IDs` prompt context from `passage["entities"]`; an empty or missing list there forces phantom-ID emissions in the prose call (the model invents short IDs from the passage_id text). A missing source passage is a structural failure (R-6.2 application order guarantees the source exists first) and must raise rather than silently fall back to an empty list.
 
 **Violations:**
 
@@ -536,6 +541,8 @@ R-6.5. Residue passages and their companion residue beats inherit the `entities`
 | Phase 6 creates a commit beat | Narrative-beat creation forbidden post-SEED | R-6.4 |
 | Phase 6 mutates an existing `belongs_to` edge | Should only add passage-layer and structural-beat nodes | R-6.4 |
 | Residue passage emits phantom entity IDs (e.g. `clara` instead of `character::clara_yu`) at FILL | Residue created with empty `entities` instead of inheriting from target | R-6.5 |
+| Variant passage emits phantom entity IDs at FILL | Variant created with empty `entities` instead of inheriting from base | R-6.5 |
+| Synthesized passage created against a missing source — empty `entities` propagated silently | R-6.2 application order violated; helper used silent fallback `[]` instead of raising | R-6.5 |
 
 ### Output Contract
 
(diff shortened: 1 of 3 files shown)


pvliesdonk:
Addressed claude-review nits in [5f49ce3](https://github.com/pvliesdonk/questfoundry/commit/5f49ce39):

- **Multi-paragraph docstring on `_create_variant_passage`** — trimmed back to one line per CLAUDE.md "Never write multi-paragraph docstrings". The rationale already lives on `_inherited_passage_entities`'s docstring.
- **ValueError message imprecision** — generalised "base passages" → "source passages (residue targets and variant bases)" so the diagnostic is precise from either call site.

Skipped the spec-doc wording observation per claude-review's "no change required, but it could be clarified" — the violation row reads as describing the bug to catch (which is the table's purpose).

Triggering gemini below for a fresh review on the latest commit.

PR #1467, opened 2026-04-27
fix(providers): switch Gemini safety_settings from BLOCK_NONE to OFF (#1466)

pvliesdonk:
## Summary

Closes #1466.

PR #1465 set Gemini \`safety_settings\` to \`BLOCK_NONE\` for all five harm categories. Murder3 re-run at 16:51 (after #1465 merged at 16:49) still hit the same \`PROHIBITED_CONTENT\` block on \`fill_phase1_expand\`:

\`\`\`
google_safety_settings_injected categories=5      ← #1465 fix is applied
chat_model_created provider=google model=gemini-2.5-pro
...
WARNING Gemini produced an empty response.
        Feedback: block_reason=<BlockedReason.PROHIBITED_CONTENT>
                  safety_ratings=None
WARNING fill_llm_validation_fail (×3)
WARNING batch_item_failed
\`\`\`

The blocked content was a beat-blueprint prompt for 8 cozy-mystery passages mentioning a blackmail ledger, an undercover journalist, and "who is protecting the killer" — standard genre tropes, nothing that should hard-block.

## Diagnosis

\`block_reason=PROHIBITED_CONTENT\` lives on \`prompt_feedback\` (with \`safety_ratings=None\`), meaning the block fires at the **input prefilter**, before generation. The output-side filter governed by \`safety_settings\` thresholds never runs.

| Threshold | Effect |
|---|---|
| `BLOCK_NONE` | "never blocks" but only governs the response filter |
| `OFF` | "turns off the safety filter" — disables filtering at **both** input and output layers |

We picked `BLOCK_NONE` in #1465 thinking it was the most permissive. `OFF` is strictly more permissive and is the documented threshold to bypass the prompt prefilter for creative-fiction tooling.

## Fix

Swap `HarmBlockThreshold.BLOCK_NONE` → `HarmBlockThreshold.OFF` for all five categories in `_default_google_safety_settings`. The `OFF` enum is exposed by `langchain-google-genai>=2.0` (pyproject pin satisfied), forwarded unchanged via `_format_safety_settings` to the underlying `google.genai` SDK.

## Verification

\`\`\`bash
$ GOOGLE_API_KEY=test-key uv run python -c \"
from questfoundry.providers.factory import create_chat_model
m = create_chat_model('google', 'gemini-2.5-pro')
for k, v in (m.safety_settings or {}).items():
    print(f'{k.name} -> {v.name}')
\"
HARM_CATEGORY_HARASSMENT -> OFF
[...]

Issue #1466, opened 2026-04-27
PROVIDERS: Gemini BLOCK_NONE insufficient — input prefilter still rejects creative content; switch to OFF


pvliesdonk:
## Summary

PR #1465 set Gemini \`safety_settings\` to \`BLOCK_NONE\` for all five harm categories. murder3 re-run at 16:51 still hits the same \`PROHIBITED_CONTENT\` block on \`fill_phase1_expand\`:

\`\`\`
google_safety_settings_injected categories=5  ← fix is applied
chat_model_created provider=google model=gemini-2.5-pro
...
WARNING Gemini produced an empty response.
        Feedback: block_reason=<BlockedReason.PROHIBITED_CONTENT: 'PROHIBITED_CONTENT'>
                  block_reason_message=None safety_ratings=None
WARNING fill_llm_validation_fail (×3 attempts)
WARNING batch_item_failed
\`\`\`

The blocked content is **standard cozy-mystery prose** (a beat-blueprint prompt for 8 passages mentioning a blackmail ledger, an undercover journalist, and \"who's protecting the killer\"). Nothing here is structurally outside the Gemini ToS for fiction.

## Diagnosis

\`block_reason=PROHIBITED_CONTENT\` lives on \`prompt_feedback\`, not \`candidates[0]\` — so the block fires at the **input prefilter**, before generation. The output filter (where \`safety_settings\` thresholds apply) never even runs.

Per Google's safety-settings docs, the threshold values differ:

| Threshold | Effect |
|---|---|
| `BLOCK_NONE` | \"Never blocks\" — but only governs the **output** filter. |
| `OFF` | \"Turns off the safety filter\" — disables filtering at both **input and output** layers per the API's intended semantics. |

We picked `BLOCK_NONE` in #1465 thinking it was the most permissive option. `OFF` is strictly more permissive and is the documented way to bypass the prompt prefilter.

## Fix

In `_default_google_safety_settings`, swap `HarmBlockThreshold.BLOCK_NONE` → `HarmBlockThreshold.OFF` for all five categories. The \`OFF\` enum value is exposed by `langchain-google-genai>=2.0` (pyproject pin satisfied), and forwarded unchanged via \`_format_safety_settings\` to the underlying \`google.genai\` SDK.

Update tests to assert the threshold is \`OFF\` instead of \`BLOCK_NONE\`.

## Verification

\`\`\`bash
$ GOOGLE_API_KEY=test-key uv run python -c \"
[...]

PR #1469, opened 2026-04-27
fix(fill): escalate batch_llm_calls failures per-passage instead of silent-drop (#1468)

### LINKED ISSUE #1468: FILL: silently drops failed batch items — escalate visibly when all retries exhaust
## Summary

All three batched LLM phases in FILL (\`_phase_1_expand\`, \`_phase_2_review\`, \`_phase_3_revision\`) call \`batch_llm_calls(...)\` and discard the returned errors:

\`\`\`python
results, total_llm_calls, total_tokens, _errors = await batch_llm_calls(...)
                                          ^^^^^^^
for blueprints in results:
    if blueprints is None:
        continue   # silently skipped
\`\`\`

The leading underscore signals "intentionally unused" — but this is the silent-fallback anti-pattern from \`.gemini/styleguide.md\`. When a chunk fails all 3 retries (e.g. Gemini returns empty due to safety blocks, or transient errors not recoverable by the connectivity-retry loop), the affected passages get *no* blueprint / no review flag / no revision — and FILL continues as if nothing happened.

## Reproducer (murder3)

PR #1465 → #1467 saga: \`fill_phase1_expand\` returned \`PROHIBITED_CONTENT\` for an 8-passage batch. After 3 retries, \`batch_item_failed\` logged once at WARNING and the run continued with 8 passages missing their expand-phase blueprint. Downstream Phase 1 prose generation produced lower-quality prose for those passages without anyone knowing which ones.

The \`safety_settings=OFF\` fix in #1467 unblocked the prefilter for that specific batch, but the silent-drop pattern remains. Future projects on different content / providers / network conditions will hit it again, with no visibility.

## Fix

For each FILL \`batch_llm_calls\` call site:

1. Capture the errors list (drop the leading underscore).
2. After the results loop, iterate \`errors\` and for each \`(idx, exc)\` pair look up the affected items from the input \`chunks\`/\`batches\`/\`passage_items\` list.
3. Append a \`FillEscalation\` per affected passage with \`kind\` matching the phase (\`expand_batch_failed\` / \`review_batch_failed\` / \`revision_failed\`), the passage_id, a detail describing what's missing downstream, and \`upstream_stage="FILL"\`.

The existing exit-time \`_check_escalations\` machinery (\`fill.py\` Output Contract) already raises a \`FillContractError\` listing all escalations, so the user sees the full set at stage exit instead of having to dig through DEBUG logs.

## Out of scope

- The deeper concern of \"why do prose-call retries fail and what should be done about it\" (provider fallback, prompt rewording, batch-size tuning) is a separate design question. This issue is purely about visibility — making failures observable so the user can decide what to do.
- Connectivity retries are already handled by \`batch_llm_calls\`'s built-in retry loop; this fix kicks in only for non-retryable failures that exhaust the inner repair loop.

## Verification

\`\`\`bash
$ uv run pytest tests/unit/test_fill_stage.py -k 'batch_failed or escalation' -x -q
# new tests pass
\`\`\`

After fix, a forced batch failure (mock \`_fill_llm_call\` to raise) should produce \`FillEscalation\` entries visible in \`stage.execute()\`'s \`escalations\` field — and trigger \`FillContractError\` with the affected passage IDs at end of stage.

## Reference

- Anti-pattern: \`.gemini/styleguide.md\` §4 \"Silent fallbacks that hide bugs\"
- CLAUDE.md \"silent degradation of story structure constraints\"
- Companion: #1466 / PR #1467 (Gemini \`safety_settings=OFF\`); the silent drop is what made #1466 hard to diagnose initially

pvliesdonk:
## Summary

Closes #1468.

All three batched FILL phases (\`_phase_1a_expand\`, \`_phase_2_review\`, \`_phase_3_revision\`) bound the \`batch_llm_calls\` errors list to \`_errors\` (intentionally-unused convention) and discarded it. When a chunk's retries exhausted (provider safety block, transient errors past the connectivity-retry loop), \`results[idx]\` became \`None\`, the \`if x is None: continue\` loop skipped it silently, and FILL continued reporting success while passages were missing their blueprint / review / revision.

This is the diagnostic backstop for #1466 / PR #1467: even when a future project / provider / content combination trips a filter we can't bypass via \`safety_settings\`, the user gets a clear signal about which passages need attention rather than a silently-degraded run.

Per \`.gemini/styleguide.md\` §4 \"silent fallbacks that hide bugs\": the failures are now surfaced as \`FillEscalation\` entries that flow through the existing exit-time \`FillContractError\` machinery, so the user sees the full set of affected passages at stage exit.

## Three new escalation kinds

\`FillEscalation.kind\` extended (in \`src/questfoundry/models/fill.py\`):

| Kind | Phase | What's missing downstream |
|---|---|---|
| \`expand_batch_failed\` | 1a | No expand-phase blueprint; Phase 1 prose generated without the blueprint context |
| \`review_batch_failed\` | 2 | No review flags; flat_prose / blueprint_bleed / continuity issues silently un-flagged for revision |
| \`revision_failed\` | 3 | Original prose retained with the prior phase-2 flags unaddressed |

Each escalation includes the passage_id, the original exception type/message, and a detail string explaining the downstream impact.

## Implementation

Per call site in \`fill.py\`:

\`\`\`python
results, total_llm_calls, total_tokens, errors = await batch_llm_calls(...)

for idx, exc in errors:
    chunk_ids, _arc_id = chunks[idx]   # or batches[idx], or passage_items[idx][0]
    for affected_pid in chunk_ids:
        full_pid = (
            affected_pid
            if affected_pid.startswith(\"passage::\")
            else f\"passage::{affected_pid}\"
        )
        self._escalations.append(
            FillEscalation(kind=..., passage_id=full_pid, detail=..., upstream_stage=\"FILL\")
        )
\`\`\`

The shape is consistent across the three call sites; only the chunk-shape and escalation kind differ. Phase 1a and Phase 2 chunks are lists of passage IDs (one escalation per affected passage); Phase 3 \`passage_items\` is one passage per item (one escalation per failed item).

## Tests

\`TestFillBatchFailureEscalations\` (in \`tests/unit/test_fill_stage.py\`) adds:

- \`test_three_new_escalation_kinds_accepted\` — Pydantic Literal accepts all three new kinds (\`expand_batch_failed\`, \`review_batch_failed\`, \`revision_failed\`)
- \`test_expand_batch_failure_escalates_each_passage_in_chunk\` — patches \`batch_llm_calls\` to return errors for every chunk, calls \`_phase_1a_expand\` on a real graph, asserts one escalation per affected passage with the right kind / passage_id / detail / upstream_stage

\`\`\`
$ uv run pytest tests/unit/test_fill_stage.py -x -q
100 passed
$ uv run pytest tests/unit/ -k fill -x -q
422 passed
\`\`\`

## Out of scope

- The deeper concern of \"why do prose-call retries fail and what should be done about it\" (provider fallback, prompt rewording, batch-size tuning) is a separate design question. This PR is purely about visibility — making failures observable so the user can decide what to do.
- Connectivity retries are already handled by \`batch_llm_calls\`'s built-in retry loop; this fix kicks in only for non-retryable failures that exhaust the inner repair loop.

## Test plan

- [x] Targeted fill tests pass (100 in test_fill_stage; 422 across all fill files)
- [x] FillEscalation Pydantic schema extended without breaking existing escalation kinds

### DIFF
--- a/src/questfoundry/models/fill.py
+++ b/src/questfoundry/models/fill.py
@@ -284,6 +284,9 @@ class FillEscalation(BaseModel):
         "voice_research_failed",
         "blueprint_validation_failed",
         "entity_extract_failed",
+        "expand_batch_failed",
+        "review_batch_failed",
+        "revision_failed",
     ] = Field(description="What kind of escalation this is.")
     passage_id: str = Field(
         description="Passage where the escalation was raised. Empty string if not passage-scoped."
--- a/src/questfoundry/pipeline/stages/fill.py
+++ b/src/questfoundry/pipeline/stages/fill.py
@@ -1312,11 +1312,9 @@ class FillStage:
             on_connectivity_error=self._on_connectivity_error,
         )
 
-        # Escalate per-passage when an entire chunk's retries exhausted
-        # (e.g. provider safety block, transient errors past the connectivity
-        # retry loop). Per `.gemini/styleguide.md` §4 "silent fallbacks that
-        # hide bugs": surface which passages have NO blueprint so the user
-        # sees the gap at stage exit instead of guessing from prose quality.
+        # Escalate per-passage when a chunk's retries exhausted: each
+        # passage in the failed chunk gets no blueprint, downstream Phase 1
+        # prose is generated without it.
         for idx, exc in errors:
             chunk_ids, _arc_id = chunks[idx]
             for affected_pid in chunk_ids:
@@ -1324,6 +1322,11 @@ class FillStage:
                     affected_pid
                     if affected_pid.startswith("passage::")
                     else f"passage::{affected_pid}"
+                )
+                log.warning(
+                    "expand_batch_failed_escalated",
+                    passage_id=full_pid,
+                    error=f"{type(exc).__name__}: {exc}",
                 )
                 self._escalations.append(
                     FillEscalation(
@@ -1901,6 +1904,11 @@ class FillStage:
                     if affected_pid.startswith("passage::")
                     else f"passage::{affected_pid}"
                 )
+                log.warning(
+                    "review_batch_failed_escalated",
+                    passage_id=full_pid,
+                    error=f"{type(exc).__name__}: {exc}",
+                )
                 self._escalations.append(
                     FillEscalation(
                         kind="review_batch_failed",
@@ -2097,11 +2105,17 @@ class FillStage:
         # Per-passage escalation when a revision call's retries exhausted —
         # the passage keeps its old (flagged-as-broken) prose with no fix.
         for idx, exc in errors:
-            affected_pid, _flags = passage_items[idx]
+            raw_pid, _flags = passage_items[idx]
+            full_pid = raw_pid if raw_pid.startswith("passage::") else f"passage::{raw_pid}"
+            log.warning(
+                "revision_failed_escalated",
+                passage_id=full_pid,
+                error=f"{type(exc).__name__}: {exc}",
+            )
             self._escalations.append(
                 FillEscalation(
                     kind="revision_failed",
-                    passage_id=affected_pid,
+                    passage_id=full_pid,
                     detail=(
                         f"fill_phase3_revision call failed after retries "
                         f"({type(exc).__name__}: {exc}). Passage retains its "
--- a/src/questfoundry/pipeline/stages/fill.py
+++ b/src/questfoundry/pipeline/stages/fill.py
@@ -1305,12 +1305,40 @@ class FillStage:
 
             return [bp.model_dump() for bp in output.blueprints], llm_calls, tokens
 
-        results, total_llm_calls, total_tokens, _errors = await batch_llm_calls(
+        results, total_llm_calls, total_tokens, errors = await batch_llm_calls(
             chunks,
             _expand_chunk,
             self._max_concurrency,
             on_connectivity_error=self._on_connectivity_error,
         )
+
+        # Escalate per-passage when an entire chunk's retries exhausted
+        # (e.g. provider safety block, transient errors past the connectivity
+        # retry loop). Per `.gemini/styleguide.md` §4 "silent fallbacks that
+        # hide bugs": surface which passages have NO blueprint so the user
+        # sees the gap at stage exit instead of guessing from prose quality.
+        for idx, exc in errors:
+            chunk_ids, _arc_id = chunks[idx]
+            for affected_pid in chunk_ids:
+                full_pid = (
+                    affected_pid
+                    if affected_pid.startswith("passage::")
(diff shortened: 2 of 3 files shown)


pvliesdonk:
Addressed all 5 claude-review findings in [ef61774](https://github.com/pvliesdonk/questfoundry/commit/ef617742):

- **#1 Phase 3 prefix inconsistency (blocking)** — applied the same `full_pid = ... if .startswith('passage::') else f'passage::{...}'` pattern across all three phases.
- **#4 Dead `chunk_passage_ids` variable (blocking)** — removed; was a placeholder with a deferred-work comment that violated the no-deferral-without-issue rule.
- **#2 Missing fill-level log calls** — added `log.warning('expand_batch_failed_escalated' / 'review_batch_failed_escalated' / 'revision_failed_escalated')` per escalation, matching the pattern at the existing `blueprint_validation_failed` and `entity_extract_failed` sites.
- **#3 `.gemini/styleguide.md` citation in source comment** — trimmed the 5-line block to one line; the WHY is already in the PR description and the FillContractError machinery.
- **#5 Missing Phase 2 + Phase 3 behavioural tests** (also caused the codecov/patch fail at 53.33%) — added `test_review_batch_failure_escalates_each_passage_in_batch` and `test_revision_failure_escalates_each_failed_passage`. Each exercises its phase's specific batch shape (Phase 2 `list[list[str]]`; Phase 3 `list[tuple[str, list[dict]]]`) so the per-shape unpacking is pinned.

`uv run pytest tests/unit/test_fill_stage.py -x -q` → 102 passed (was 100). mypy + ruff + pre-commit clean. Codecov should rise above the 80% threshold with the two new integration tests now covering the previously-uncovered Phase 2 + Phase 3 escalation lines.

Gemini hit its daily quota on this PR's first review (same as #1467 earlier today). Will not re-trigger; relying on claude-review's second pass for convergence.

pvliesdonk:
Addressed all 3 minor observations from the second-pass review in [92f4869](https://github.com/pvliesdonk/questfoundry/commit/92f4869c):

- **(a)** Collapsed each escalation comment block (2-3 lines → 1 line) per CLAUDE.md "one short line max".
- **(b)** Replaced `error=f"{type(exc).__name__}: {exc}"` with separate `exc_type=` + `exc_msg=` structured log keys per CLAUDE.md "pass structured key=value context, not f-string interpolation".
- **(c)** Added `"RuntimeError" in e.detail` assertions to Phase 2 + Phase 3 escalation tests for parity with Phase 1a.

Pure cleanup, no behaviour change. Local: ruff + mypy + 102 fill-stage tests all green.

Gemini still at daily quota — relying on claude-review's third pass for final convergence sign-off.

PR #1554, opened 2026-04-28
fix(prompts): wire SizeProfile.fully_explored into SEED dilemmas prompt (#1236)

### LINKED ISSUE #1236: fix(prompts): wire SizeProfile.fully_explored into dilemmas prompt (long stories need >=3 fully-explored)
## Context

`min_arcs_required` in `src/questfoundry/pipeline/stages/seed.py:488` is computed as `max(2, max_arcs // 4)`. For each preset:

| Preset | max_arcs | min_arcs_required | min fully-explored dilemmas (since arc_count = 2^n) |
|---|---|---|---|
| vignette | 2 | 2 | 1 |
| short | 8 | 2 | 1 |
| standard | 16 | 4 | 2 |
| long | 32 | 8 | 3 |

PR #1235 fixed the dilemmas prompt to push the LLM toward branching, but it hardcodes "at least 2" everywhere. For a `long` story this is insufficient — the run will still hit `SeedStageError` at the arc-count check, just with a higher bar.

## Scope

- Add `fully_explored` to `size_template_vars()` in `src/questfoundry/pipeline/size.py` (it's already a field on `SizeProfile` — just exposed as a template var).
- Inject `{size_fully_explored}` into the dilemmas prompt via `serialize.py` (similar to how `{size_beats_per_path}` is injected for beats).
- Replace the hardcoded "at least 2" in the dilemmas_prompt's HARD constraint and FINAL CHECK with `{size_fully_explored}` — so the prompt scales with the size profile.
- Update tests to assert injection works (`test_size.py` for the var addition, `test_serialize.py` for the prompt substitution).

## Why deferred from #1235

The immediate failure mode (all-default → 0 branching → fail-fast on standard preset) is fixed by #1235. The long-preset gap is narrower (only fires when running long-scale stories) and benefits from a separate review pass focused on the size-vars wiring, which is mechanically distinct from the prompt restructure.

Caught by `claude[bot]` review on PR #1235:
> `min_arcs_required` is computed in `seed.py:488` as `max(2, max_arcs // 4)`. For the `long` size profile (`max_arcs=32`), that's **8**. Arc count = `2^n` where `n` = fully-explored dilemmas, so passing requires `n ≥ 3`. The prompt now hardcodes `at least 2` everywhere.

## Acceptance

- `size_template_vars()` returns `size_fully_explored` for every preset
- `serialize.py` substitutes `{size_fully_explored}` in the dilemmas prompt
- Hardcoded `at least 2` in the dilemmas section is replaced with the placeholder
- Manual verification: re-run `projects/test-new/` against `qwen3:4b` with `story_size: long` and SEED reaches GROW

pvliesdonk:
## Summary

The dilemmas prompt's "at least 2 dilemmas MUST be fully explored" hard
constraint was hardcoded everywhere. PR #1235 fixed the prompt structure
but left the magic 2 — fine for short/medium, insufficient for `long`.
With `max_arcs=32`, validation requires `min_arcs_required=8` arcs
(`max(2, max_arcs // 4)`), which means at least 3 fully-explored
dilemmas (`arc_count = 2^n`). The prompt asking for 2 falls short and
the run hits `SeedStageError`.

This wires `SizeProfile.fully_explored` (already a field — just not
exposed) through `size_template_vars()` and into:

- `serialize_seed_sections.yaml` `dilemmas_prompt` — HARD CONSTRAINT + FINAL CHECK
- `discuss_seed.yaml` — MINIMUM BRANCHING REQUIREMENT

`serialize.py::serialize_seed_as_function` now applies the
`{size_fully_explored}` substitution alongside the existing beat-range
substitutions. The `discuss_seed.yaml` substitution rides on the
existing `**size_template_vars(...)` `.format()` call in
`agents/prompts.py::get_seed_discuss_prompt` — no extra wiring needed.

### Profile values (TARGET, always >= validation MIN)

| preset | max_arcs | min required | `size_fully_explored` |
|--------|---------:|-------------:|----------------------:|
| micro  |        2 |            1 |                     1 |
| short  |        8 |            1 |                     3 |
| medium |       16 |            2 |                     4 |
| long   |       32 |            3 |                     5 |

The profile target is always `>=` the validation minimum, so the
prompt asks for slightly more than what's strictly needed — which is
the right direction (LLM tries for more, falls short of TARGET, but
still satisfies validation).

Closes #1236

## Test plan

- [x] `uv run mypy src/questfoundry/pipeline/size.py src/questfoundry/agents/serialize.py` → Success
- [x] `uv run ruff check` → All checks passed
- [x] `uv run pytest tests/unit/test_size.py tests/unit/test_serialize.py` → 142 passed
- [x] New test `test_size_fully_explored_scales_with_preset` asserts the new key for all 4 presets
- [x] New test `test_dilemmas_prompt_size_fully_explored_placeholder_present` asserts no hardcoded "at least 2" / "less than 2" / "fewer than 2" strings survive in the dilemmas prompt

### DIFF
--- a/prompts/templates/discuss_seed.yaml
+++ b/prompts/templates/discuss_seed.yaml
@@ -75,10 +75,11 @@ system: |
 
   ## MINIMUM BRANCHING REQUIREMENT (CRITICAL)
 
-  **At least 2 dilemmas MUST have BOTH answers explored.** Without at least
-  one Y-fork, POLISH Phase 4c produces 0 choice edges (see seed.md §Overview)
-  and the story collapses to a linear track. The downstream pipeline assumes
-  a real branching structure.
+  **At least {size_fully_explored} dilemmas MUST have BOTH answers explored.**
+  Without enough Y-forks, POLISH Phase 4c produces too few choice edges (see
+  seed.md §Overview) and the story collapses toward a linear track. The
+  downstream pipeline assumes real branching that scales with story size:
+  small stories need at least one fork; longer stories need more.
 
   GOOD (3 dilemmas, 2 fully explored):
   - `dilemma::keeper_honest_or_hiding`: explored = `[honest, hiding]` — both answers
@@ -89,9 +90,10 @@ system: |
   - All dilemmas: explored = `[canonical_answer_only]` → linear story, 0 choices
   - Only 1 dilemma fully explored → story has only one fork; POLISH Phase 4c produces ≤2 choices
 
-  Pick at least 2 dilemmas where exploring both answers serves the story (per the
-  Identity-defining / Genuine dilemma / Distinct content tests above). Beyond
-  that minimum, explore more if compelling — the size preset bounds the upper end.
+  Pick at least {size_fully_explored} dilemmas where exploring both answers serves
+  the story (per the Identity-defining / Genuine dilemma / Distinct content tests
+  above). Beyond that minimum, explore more if compelling — the size preset bounds
+  the upper end.
 
   ### 3. Create Paths from Explored Answers
 
--- a/prompts/templates/discuss_seed.yaml
+++ b/prompts/templates/discuss_seed.yaml
@@ -81,7 +81,8 @@ system: |
   downstream pipeline assumes real branching that scales with story size:
   small stories need at least one fork; longer stories need more.
 
-  GOOD (3 dilemmas, 2 fully explored):
+  GOOD (structure example — showing what "fully explored" vs "canonical only"
+  looks like; your story needs at least {size_fully_explored} fully explored):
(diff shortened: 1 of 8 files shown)


pvliesdonk:
Filed #1555 to track the stale 'at least 2 dilemmas' error message in seed.py:527 — out of scope for this PR (user-facing recovery guidance vs prompt-side prevention) but worth fixing.

pvliesdonk:
Filed #1556 to track the round-6 `discuss_seed.yaml:92` BAD-example contradiction ("only 1 dilemma fully explored" labelled BAD vs micro's `size_fully_explored=1`). Pre-existing line not touched by this PR's commits; runtime is unaffected since over-targeting on micro still validates. Flipping ready.

